PR #2, opened 2018-10-04
Added the option to set how many hours to show in graph

xiidoz:


kalkih:
Awesome, this is great! 👍🏼

PR #16, opened 2018-11-30
Update README.md

vMeph:
change mini-graph-card-bundle.js to mini-graph-card.js

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,13 +9,13 @@ The card works with entities from within the **sensor** domain and displays the 
 
 ### Simple install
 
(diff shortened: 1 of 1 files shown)


kalkih:
The readme is actually correct, while you technically still can use the source file, you are meant to use the bundle as the instructions describe..

The bundle is a compiled version of the source code, with external dependencies bundled in.
This resolves in the card loading faster, improves compatibility and make it work offline.

vMeph:
ohh ok sounds cool

PR #20, opened 2019-01-14
🛠️ use local lit

iantrich:


kalkih:
Waiting a bit to merge this since it's not compatible with HA < 0.80.0.
Discussed further here kalkih/mini-media-player#58

PR #25, opened 2019-01-26
add `group` option to disable padding and box-shadow

iantrich:


iantrich:
ugh, how did I not check for that first?! been missing out on graph animations :) I'll get it sorted. Thanks.

PR #26, opened 2019-01-26
add `group` option to disable padding and box-shadow

iantrich:


iantrich:
Left the changelog and readme updates to you as you have more changes obviously :)

mini-graph-card.js
```diff
@@ -324,6 +325,9 @@ class MiniGraphCard extends LitElement {
           padding: 16px;
           position: relative;
         }
+        ha-card[group] {
+          box-shadow: none;
```
kalkih: Should set padding: 0; here also

mini-graph-card.js
```diff
@@ -151,6 +151,7 @@ class MiniGraphCard extends LitElement {
       ${this._style()}
       <ha-card
         class='flex'
+        ?bg=${config.background} ?group=${config.group}
```
kalkih: `bg` attribute is not necessary

iantrich: that's what i get for being a sloppy copy/paster

iantrich:
ooo I really like that animate

iantrich:
multiple entities, not yet. Been working to get this looking smooth on my finance card with just a net worth graph. working great so far

Issue #83, opened 2019-04-21
Best way to show long term data?


rdehuyss:
I've noticed that HomeAssistant gets unresponsive if showing history data for about two weeks using the default sqlite database.

Is it possible to do some throttling when asking for the history data?

I have 4 mini-graph-cards showing history like this:

```
          - entities:
              - entity: sensor.floor_2_kwh_verlichting_total
                name: VERLICHTING
            font_size: 75
            hours_to_show: 336
            icon: 'mdi:flash'
            line_width: 8
            points_per_hour: 0.041667
            type: 'custom:mini-graph-card'
```

But the result is that my homeassistant gets unresponsive and goes to 100% CPU due to the history api calls.

rdehuyss:
I just noticed that the network calls also queue... i.e. a new request is launched before the previous request is finished. My guess would be to only request a new history api call if the previous one completed. See the attached screenshot:

![image](https://user-images.githubusercontent.com/567842/56467101-9b5cfa80-641a-11e9-860c-76d910b0e236.png)

kalkih:
Hi,
Yes, I've seen some users having issues with the history API before, was discussed quite a bit in the mini-graph-card thread in the HA community forum but without any real conclusion I think.
Changing to a different database did seem to help in some cases but history requests still seemed to be relatively slow.

Not sure if history performance has been improved in later versions of HA but I seem to load stuff pretty quick nowadays, I'm running MariaDB however.
Before, loading the logbook or the history tab would take like 30+ sec.

Queueing requests is not a bad idea and would certainly help when the entity(ies) updates very frequently, even better would be to add some sort of caching, where we only fetch new history since our last request.

If you want a quick solution, I guess you could create a template sensor for the sensor and make sure it only updates every x minutes.

bramkragten:
I can think of 2 solutions:

Instead of updating the history on every state update, you could update it on a time interval if the state is changed, this could be a setting.

And on updating only get the new data instead of all data, store the last fetch timestamp and use that as start for the history api, then concat the new data with the old.

I could look into it in the next weeks or so if you think it is a good idea.

kalkih:
> And on updating only get the new data instead of all data, store the last fetch timestamp and use that as start for the history api, then concat the new data with the old.

This would be great, basically what I had in mind as well.

Not sure if the extra "rate limit" option would even be necessary if we implement the above solution. Could possibly still be useful in low performing setups?

[...]

PR #86, opened 2019-05-13
Optimize fetching history

bramkragten:
- Don't fetch history while we are waiting for a fetch
- Add config `update_interval` to update only after a number of seconds instead of on every state change
- Only fetch the history from the last time we fetched instead of all

Fixes https://github.com/kalkih/mini-graph-card/issues/83

My prettier/linter might have been a little too happy... Sorry, I think yours should fix it at once? Might want to share your settings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "@babel/preset-env": "^7.3.1",
     "babel-plugin-iife-wrap": "^1.1.0",
     "babel-preset-minify": "^0.5.0",
-    "eslint": "^5.3.0",
+    "eslint": "^5.15.0",
     "eslint-config-airbnb-base": "^13.1.0",
     "eslint-plugin-import": "^2.17.2",
     "rollup": "^0.66.6",
--- a/package.json
+++ b/package.json
@@ -32,9 +32,9 @@
     "@babel/preset-env": "^7.3.1",
     "babel-plugin-iife-wrap": "^1.1.0",
     "babel-preset-minify": "^0.5.0",
-    "eslint": "^5.14.0",
+    "eslint": "^5.3.0",
     "eslint-config-airbnb-base": "^13.1.0",
-    "eslint-plugin-import": "^2.16.0",
+    "eslint-plugin-import": "^2.17.2",
     "rollup": "^0.66.6",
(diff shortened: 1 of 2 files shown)


kalkih:
Wow, that was quick! ⚡️

There's an eslint config in the base of the repo & an `npm run lint` script.
The eslint version you changed to doesn't seem to play nice with my vscode however.
And yeah, the eslint config doesn't work that well with lit-html and the template literals...

Looks good!

bramkragten:
Also created a branch that uses `localStorage` for caching, but I don't think that should be in this card. It does speed up my energy monitor graph that is updating every second though ;-)

<https://github.com/bramkragten/mini-graph-card/tree/localstorage>

kalkih:
Hmmm, that's pretty awesome, speeds up thing quite a lot in many use cases. I don't see an issue with that being merged into the card in the future, (in another PR ofc).

Btw, not sure if this will cause issues with duplicate data entries for us, but I believe it could.
The history endpoint seems to always return one history entry prior to the `start_time`, so if you make a request "from the future" you seem to get the latest history entry even if it's "from the past".
So in our case we seem to always get an extra entry back with the timestamp of the `start_time` provided in the request.

`http://hyperion:8123/api/history/period/2019-05-13T14:29:51.076Z?filter_entity_id=sensor.load_1m&end_time=2019-05-13T14:30:21.048Z`

Screenshot 2019-05-13 at 16 30 41

I haven't looked into it much, but I guess we will end up with a duplicate history entry for every new request we make.

> And we might have to add purging, it now only adds data and never removes old data...

Haha yeah, I guess that makes sense 😆

kalkih:
Hope you didn't mind me changing the formatting of the template literals.

bramkragten:
> Hope you didn't mind me changing the formatting of the template literals.

No not at all :-) I messed up your formatting in the first place

bramkragten:
Will look into purging and the extra data tomorrow 👍

bramkragten:
Prettier doesn't agree with you on the template literals 😞

kalkih:
Awesome, will take a closer look tomorrow but looks good!

kalkih:
Closing in favor of #88.

PR #88, opened 2019-05-14
Use localStorage for caching history data

bramkragten:
Just dropping it here for now. Works ⚡️ fast for me, ~~especially because I render with the cached data before the first fetch is done (as that can take some time)...~~

Things we should think about:
- ~~I store the history based on the `entity_id` if a user would use an entity in more than one graph with different `hours_to_show` settings, it would mess up the data...~~ I now store the `hours_to_show` setting, that means that 2 graphs with the same entity and different `hours_to_show` will not use the cache.
- ~~Increasing the `hours_to_show` setting will not show instant, as it will not fetch new data before the last time. It will not purge however so eventually you will get the right data.~~ I now store the `hours_to_show` setting, after changing it, it will fetch everything.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -3,6 +3,7 @@ import Graph from './graph';
 import style from './style';
 import {
   URL_DOCS,
+  HISTORY_STORAGE,
   FONT_SIZE,
   FONT_SIZE_HEADER,
   MAX_BARS,
@@ -35,7 +36,6 @@ class MiniGraphCard extends LitElement {
     this.gradient = [];
     this.tooltip = {};
     this.updateQueue = [];
-    this.history = [];
     this.updating = false;
     this.stateChanged = false;
   }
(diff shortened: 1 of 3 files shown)


bramkragten:
Also interesting... Getting multiple entities at once is way faster...
Can't explain why a single entity would take longer, but it does...

image

bramkragten:
> Does this make #86 redundant?

Yes

> Do you think we should/could make caching optional (enabled by default)? Not sure why you wouldn't want it enabled but I guess it would be nice to have the option.

We could... not sure

> Interesting, kinda expected this.
> I feel like since we mostly fetch history reactively based on entity state changes & independently of each other, I guess it shouldn't matter too much. Would be ideal for the initial requests though I guess.

Looked into this, and made PR to core to fix it <https://github.com/home-assistant/home-assistant/pull/23881>
It is in the `initial_state` and we skip that on updates now anyway :-)

src/main.js
```diff
@@ -113,28 +124,39 @@ class MiniGraphCard extends LitElement {
...
-        )
-      ));
+        ),
+      );
     }
 
     this.config = conf;
   }
 
+  firstUpdated() {
```
kalkih: What do you think about replacing firstUpdated with connectedCallback & also add disconnectedCallback?

```js
  connectedCallback() {
    super.connectedCallback();
    if (this.config.update_interval) {
      this.updateOnInterval();
      this.interval = setInterval(this.updateOnInterval, this.config.update_interval * 1000);
    }
  }

  disconnectedCallback() {
    clearInterval(this.interval)
    super.disconnectedCallback();
  }
```

That way we could prevent updating the history while in the background, and also clear the interval properly.
And I guess we would also get the initial data faster

kalkih:
> We could... not sure

We could wait and see if anyone ask for it, and then add it I think.

kalkih:
Thank you for the great work @bramkragten, I'll "fix" the weird template literal arrow function indentation and merge this.

kalkih:
Once again Bram, thank you very much for this PR, great addition to the card!

bramkragten:
No problem! Has to do it to enjoy your excellent work on this card! So thank you ;-)

PR #95, opened 2019-05-21
:bug: Catch if localStorage quota is exceeded

kalkih:
Dealing with storage quota exceeded errors, causing graph to not display.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -667,7 +667,13 @@ class MiniGraphCard extends LitElement {
         last_fetched: end,
         data: stateHistory,
       };
-      storage[HISTORY_STORAGE] = JSON.stringify(history);
+      try {
+        storage[HISTORY_STORAGE] = JSON.stringify(history);
+      } catch (e) {
+        // eslint-disable-next-line no-console
+        console.warn('mini-graph-card: Failed to cache, not enough space.');
+        storage.removeItem(HISTORY_STORAGE);
+      }
     }
 
     if (stateHistory.length === 0) return;


kalkih:
Not a very elegant solution, but since we actually never really purge old data (unless new data is fetched for the entity) I ended up just clearing out our localStorage if the storage size ever exceeds what's allowed.

The max size for a localStorage value seem to be 5-10MB in most browsers, users with more than a few graphs and/or longer `hours_to_show`, will exceed the limit pretty quick.

@bramkragten 
Any ideas on how we could improve this?

bramkragten:
We could use `lz-string` to at least be able to store a lot more?

bramkragten:
And maybe add the option to cache a graph or not, so a user can choose what to cache if it doesn’t fit

kalkih:
Sure, both great ideas!
Will merge this PR, since the issue is kinda critical, breaking the card for some issues.

Petro31:
having same issue if you need help reproducing

image

kalkih:
Thank you, but not necessary at this time.
This PR should fix the issue.

bramkragten:
OK, so I think the options are:

- using [localForage](https://github.com/localForage/localForage), so you can use indexedDB
- using `lz-string` to compress and still use localStorage with size restriction (so we can save a lot more but not everything...)
- make caching optional
- purge old cache items when cache is ful, of before (ie, cache items with long last fetch time or last fetch < start)
- using multiple?!

bramkragten:
Started with localForage (https://github.com/kalkih/mini-graph-card/pull/96) as that doesn't have a negative impact on speed and will help a lot.

PR #96, opened 2019-05-22
Move to localForage

bramkragten:
Use indexeddb where supported instead of localStorage to have more space available

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 /node_modules/
 /dist/
 package-lock.json
+src/.vscode/settings.json
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,4 @@
 /node_modules/
 /dist/
 package-lock.json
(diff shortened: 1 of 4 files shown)


kalkih:
Seems to work great, merging when I get home 👍🏼

bramkragten:
Added `lz-string` but their code is oldskool... needs some cleaning.
https://github.com/bramkragten/mini-graph-card/commit/c3d5f05508d20405ce529c900c319ff23895ea42

bramkragten:
If we want to add compressing, we do need some sort of check if data is compressed or not btw, for being backwards compatible...

kalkih:
Cool, yeah that would be an issue, but I guess we could possibly just switch store name if we switch to compression?
https://github.com/kalkih/mini-graph-card/pull/96/files#diff-1c90ff38a08209f9ebd4d05d1e43358eR25

PR #98, opened 2019-05-22
WIP: Optimize caching

bramkragten:
- use `lz-string` to compress data
- remove unused data from history
- purge stale items from cache once

Needs cleaning of `lz-string`, or external...


kalkih:
Awesome Bram, ignoring the data we don't use should help quite a bit!
Maybe we could utilize this package?
https://www.npmjs.com/package/lz-string

bramkragten:
yeah, having some problems getting that package to work with rollup... Maybe you can help?

kalkih:
Okay, will take a look tomorrow!

kalkih:
> yeah, having some problems getting that package to work with rollup... Maybe you can help?

Hmm... same, couldn't get the import to work at all 😞

kalkih:
Thinking about moving the lz-string source into another repo, and add it as a dependency instead.
What do you think?

bramkragten:
Yeah fine with me, maybe there is another good lib for it, have not really searched hard...

kalkih:
I did actually look around but didn't really find anything 😞

kalkih:
Moved it now, had some weird issues when trying to import compress/decompress directly from `main.js`, so kept the imports in utils.

bramkragten:
LGTM

kalkih:
Sorry!
Merging when I get home tmrw, or if you wanna merge it into `dev`, I think you should have permission?

kalkih:
Thanks @bramkragten, great work!

PR #115, opened 2019-07-05
First go at adding secondary y-axis, show_line and show_points

michaelblight:
At the moment I've just cloned a bunch of code, so it could be tidied depending on how you'd like to proceed. Also, I've added "show_line" and "show_points" to the entity. Given that "show_indicator" and "show_state" are already there, maybe entity should get a "show" collection that overrides the one for the card.

So far I have only tested the changes I planned, not that I haven't broken anything else, or the effect of hovering the mouse, so there's still a bit of work to do.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -338,10 +338,10 @@ class MiniGraphCard extends LitElement {
   }
 
   renderLegend() {
-    if (this.config.entities.length <= 1 || !this.config.show.legend) return;
+    if (this.visibleEntities.length <= 1 || !this.config.show.legend) return;
     return html`
       <div class="graph__legend">
-        ${this.entity.map((entity, i) => html`
+        ${this.visibleEntities.map((entity, i) => html`
           <div class="graph__legend__item"
             @click=${e => this.handlePopup(e, entity)}
             @mouseover=${() => this.setTooltip(i, -1, this.entity[i].state, 'Current')}
@@ -607,6 +607,10 @@ class MiniGraphCard extends LitElement {
     return this.config.entities[i].color || threshold.color;
   }
 
+  get visibleEntities() {
+    return this.config.entities.filter(entity => entity.show_graph !== false);
+  }
+
   intColor(inState, i) {
     const { color_thresholds, line_color } = this.config;
     const state = Number(inState) || 0;
@@ -724,7 +728,10 @@ class MiniGraphCard extends LitElement {
   }
 
   async updateEntity(entity, index, initStart, end) {
-    if (!entity || !this.updateQueue.includes(entity.entity_id)) return;
+    if (!entity
+      || !this.updateQueue.includes(entity.entity_id)
+      || this.config.entities[index].show_graph === false
+    ) return;
     let stateHistory = [];
     let start = initStart;
     let skipInitialState = false;


src/main.js
```diff
@@ -545,6 +549,16 @@ class MiniGraphCard extends LitElement {
     `;
   }
 
+  renderLabelsSecondary() {
+    if (!this.config.show.labels_secondary) return;
+    return html`
+      <div class="graph__labels flex" style="right: 0">
```
kalkih: Let's move the inline style to `style.js`, also need to get rid of the right-margin.
Class can be "graph__labels --secondary" or something so we can target it.

src/main.js
```diff
@@ -222,6 +224,7 @@ class MiniGraphCard extends LitElement {
         ?fill=${config.show.graph && config.show.fill}
         ?points=${config.show.points === 'hover'}
         ?labels=${config.show.labels === 'hover'}
+        ?labels-secondary=${config.show.labels_secondary === 'hover'}
```
kalkih: Need to se the default of `hover` in `const.js` `DEFAULT_SHOW`, similar to the default labels.

README.md
```diff
@@ -111,7 +111,10 @@ Providing options are optional, entities can be listed directly, see example bel
 | unit | string | optional | Set a custom unit of measurement, overrides `unit` set in base config.
 | show_state | boolean | optional | Display the current state.
 | show_indicator | boolean | optional | Display a color indicator next to the state, (only when more than two states are visible).
+| show_line | boolean | optional | Set to false to turn off the line. Does not impact the fill.
+| show_points | boolean | optional | Set to false to turn off the points.
 | state_adaptive_color | boolean | optional | Make the color of the state adapt to the entity color.
+| y_axis | string | optional | If 'secondary', displays using the secondary y-axis on the right.
```
kalkih: Could we make this a boolean instead? What do you think?

src/main.js
```diff
@@ -691,25 +705,53 @@ class MiniGraphCard extends LitElement {
 
     this.updateQueue = [];
 
+    if (config.show.graph) {
+      this.entity.forEach((entity, i) => {
+        if (!entity || this.Graph[i].coords.length === 0) return;
+        this.Graph[i].y_axis = config.entities[i].y_axis === 'secondary' ? 1 : 0;
+      });
+    }
```
kalkih: Do we have to do this? We can access the config setting through `entities[i].y_axis` already

src/main.js
```diff
@@ -691,25 +705,53 @@ class MiniGraphCard extends LitElement {
...
         ? config.upper_bound
-        : Math.max(...this.Graph.map(ele => ele.max)) || this.bound[1],
+        : Math.max(...this.Graph
+          .filter(ele => ele.y_axis === 0)
+          .map(ele => ele.max)) || this.bound[1],
+    ];
+
+    this.boundSecondary = [
+      config.lower_bound_secondary !== undefined
+        ? config.lower_bound_secondary
```
kalkih: Need to add `lower_bound_secondary` to the readme (same for upper)

src/main.js
```diff
@@ -691,25 +705,53 @@ class MiniGraphCard extends LitElement {
...
+        : Math.min(...this.Graph
+          .filter(ele => ele.y_axis === 1)
+          .map(ele => ele.min)) || this.boundSecondary[0],
+      config.upper_boundSecondary !== undefined
+        ? config.upper_boundSecondary
```
kalkih: Rename to `upper_bound_secondary`

src/main.js
michaelblight: Yeah this is the sort of tidying I have yet to do. I'm assuming "--secondary" will only have "right: 0; margin-right: 0px" in it.

README.md
michaelblight: I saw there was another feature request to have multiple x-axis scales (more than 2). So I wondered if people might want to do the same thing in the y-axis - multiple data series with independent scales. So you could name them "scale1", "scale2" etc. And then if they were called "primary" or "secondary" they would have the extra meaning as displaying the graph labels on the left or right. Not that I've coded that yet - so far anything other than "secondary" implies "primary". Do you think this would be worthwhile, or overkill?

src/main.js
michaelblight: Probably not - I'll try without. I'm still getting my head around the detailed structure of the code. The place where I wanted to refer to the property was using `this.Graph` and not `config.entities`, so I thought I might need to pass it in.

src/main.js
michaelblight: Oops - I started the coding as camel-case before I spotted the snake case. I missed changing this one. Also implied from this is I haven't yet tested the upper/lower bounds on the secondary access, which I will also do.

michaelblight:
I hadn't done a screenshot because it seemed a bit arrogant, but I'm more than happy to do so. It might just be a bit tricky to match the existing size and style perfectly.

I'm having a bit of trouble understanding the mechanics of how hover is working. Despite adding the relevant code for show `labels_secondary`, both labels are still being controlled by `labels`. I've not tried debugging the code to see what's going on. Is this feasible if I put all the un-minified files in the right location?

And I haven't really looked into it much, but is it possible to mock the data rather than hooking it up to HA? It would make testing different scenarios easier and more reliable.

src/main.js
michaelblight: Done

src/main.js
michaelblight: Done

src/main.js
michaelblight: Done. I created new properties of `primaryYaxisSeries` and `secondaryYaxisSeries` to tidy it up and self-document.

src/main.js
michaelblight: Done

src/main.js
michaelblight: Done

README.md
kalkih: Probably overkill, would get pretty messy I would think.
But sure, let's keep it to future proof.

src/main.js
kalkih: Nice solution! Well done!

kalkih:
If you drag the image file into the comment field here on github it will upload the image and you get a link you can reuse in the readme, kinda hacky but I prefer this way since we 1. Don't have to host the images in the repo, 2. Don't have to care about relative paths and can reuse the readme / images on other places.
nighttime-bands

> I've not tried debugging the code to see what's going on. Is this feasible if I put all the un-minified files in the right location?

If you run `npm run watch` the bundle won't be minified and should be easier to debug.


> And I haven't really looked into it much, but is it possible to mock the data rather than hooking it up to HA? It would make testing different scenarios easier and more reliable.

Haha, yes, this project is really lacking automated testing, you could create a mock history json request and feed the graph using this data instead of making API request but requires commenting out a few lines of code and adding a few other.
I've done it myself a few times when debugging stuff, not optimal... 😆

kalkih:
Whoopsie... I screwed that one up, sorry. Could you force push your local changes? I apparently don't have permission to do so.

michaelblight:
Just realised the image you dragged in is not the scaled one, so doing it again...
nighttime-bands

kalkih:
Awesome, looks great now!
I can't seem to reopen this PR, not sure why... would you mind opening a new one.

Issue #80, opened 2019-04-03
Feature request: reference lines


snis:
I would like to be able to add reference lines (Y axle) with possibility to change color and width. The line should not fill the area below and should be rendered behind the entity sensors if possible.

In a temperature graph a line for 0C/32F would make sense. Or a threshold line (calculated from a template) showing average server loads or temperature.

![graph2](https://user-images.githubusercontent.com/205149/55472213-d209d700-560b-11e9-8a82-8b235bed4318.png)

noisemaker00:
I would like this too, in particular for temperature chart (eg. 0°C reference).
How can I achieve that? Has the feature already been released?

kalkih:
No this hasn't been implemented yet.

You could technically set up a template sensor and assign it a constant value of which you want the reference line to display (`0` in your case I guess), then use that entity in the card and utilize the various show options to adjust the visuals of it, I would recommend hiding points for example.

mmiller7:
I'd also very much like this feature - I thought this would be a good fix for having a graph with fixed scale max/min (hassio's graph apparently can't) but the lack of grid-lines to glance and see what the intermediate values are without having to trace every point with the mouse is a deal-breaker.

StefanoGiu:
Interested in this... following

JJFourie:
This would indeed be very useful and appreciated.

As the graph currently does not have any grid lines (pending a submitted feature request), it is not all that easy to orient the line with respect to reference values.
In addition, it would be nice to (optionally) also make it possible to set a "legend" for the line.

So to summarize, make it possible to configure:

1)    Reference line for each value provided.
2)    Display the line description if configured (optional).
3)    Nice to have: make it possible to configure the line appearance
       -    line color
       -    thickness
       -    dashed/solid

Below: example of a graph with three reference lines, each with a description.

![RainCard-Light](https://user-images.githubusercontent.com/49686555/104130119-4fc94d80-536f-11eb-8ff1-a85fad71dd67.jpg)

RomRider:
Duplicate of #21

(let's discuss it there while I'm trying to wrap my head around the codebase :))

PR #116, opened 2019-07-07
Secondary y-axis code finished

michaelblight:
I finally worked out how hover is working! I put `--primary` on the main y-axis to differentiate the hiding/showing of the two.


kalkih:
Looks great, well executed!
Will merge now, might do a new release in a day or two!

Thanks for your contribution! 🎉

michaelblight:
No worries. While I'm on a roll, is there anything else you'd like help with? You've tagged a couple of things with "help wanted". Shall I just pick one or do you have a preference? (I'm going to bed now as it's nearly midnight here and work looms, but I'll check back tomorrow).

kalkih:
Yes, that would be awesome, any of the issues/feature requests really, doesn't matter to me, pick one that you find interesting. 🙂

I'll add you as a collaborator, that way you can work directly against this repo if you'd like, just create a new branch from the dev branch for each feature and then a PR back to the dev branch.

michaelblight:
I just realised I forgot to remove redundant code - main.js lines 721-726. They just need to be deleted, which I can do tonight.

Separately, I tried the dark theme you use, and found the fill with 6 entities graphed was getting a bit over-the-top. So in my local version I've also provided the option to turn it off at an entity level. This probably goes a fair way to handling issue #80. I've got some thoughts on some of the other issues raised too - are you happy for me to just add comments to collaborate? I don't want to throw you under the bus or cause the component to go in a direction you're not happy with.

There's now a lot of `show_*` properties - I will look at the feasibility of moving these to a `show` collection for the entity. I will try to include all the show properties at the card level, so `card.show` is the default for all entities, and `entity.show` can override some/all for a specific entity. Then you can see what you think of it. In my mind it would simplify the usage a bit.

michaelblight:
Damn, I screwed up, sorry. I forgot to create a new branch and I pushed a couple of commits straight to dev. I hope you don't mind.

The first commit was to remove the redundant code mentioned above, which brought with it the introduction of `show_fill`. The second was to fix a bug I introduced when setting the constant for `labels_secondary`. Once I did this, the default became 'hover', and if you didn't have any secondary y-axis data, hovering showed 'infinity' in the labels. I tested these pretty thoroughly, but sorry once again. Maybe it would be better if I went back to forking the repo.

kalkih:
No worries, I don't mind doing small commits directly to dev.

michaelblight:
It looks like something went a bit wrong with the merge of README. The image I added is still pointing to the images folder which was removed, and therefore causes a broken link. However, GitHub thinks there's no difference between my fork and your dev or master branches.

kalkih:
Strange, if the image link is the only thing wrong, do a quick & dirty commit right to master with the correct link if you'd want. 😉

PR #124, opened 2019-07-15
Fix bar width when hiding some entities

caphm:
In bar graph mode, when hiding the graph for some entities via `show_graph: false` the width of the remaining graph bars was not adjusted properly.


kalkih:
Awesome, Thanks! 🙏

Issue #119, opened 2019-07-12
lower_bound not calculated correctly when a hidden entity exists


cicero200272:
When I use two entities and suppress the graph of the second one, the lower_bound for the y-axis is set to 0. If delete the second entity, the lower_bound is calculated correctly.

```
- type: custom:mini-graph-card
  entities: 
    - entity: sensor.temp1
    - entity: sensor.temp2
      show_state: false
      show_graph: false
```

![graph2](https://user-images.githubusercontent.com/15817602/61133003-db34fa80-a4bc-11e9-98a5-5af8b9803a46.png)

```
- type: custom:mini-graph-card
  entities: 
    - entity: sensor.temp1
```

![graph3](https://user-images.githubusercontent.com/15817602/61133100-0ae40280-a4bd-11e9-8ff3-73bbe62e9a1a.png)

kalkih:
Fixed in latest release, thanks for letting us know about the issue!

PR #130, opened 2019-08-06
Fixed min/max when graph is hidden

michaelblight:
This is intended to fix issue #119, but you might want to do it differently. It was caused because setting the min/max was not honouring `show_graph=false`. I realised there was a problem with using `visibleEntities` as an index into `this.Graph` because one is a filtered view, so the indexes are wrong. This is why I created a new `index` property on `this.config.entities` so a filtered list can use it to index into `this.Graph` properly. What do you think? 

This uncovered another issue where putting `show_graph: false` on the first entity will still result in its value being used for the current value, and it also won't show the extrema then. I haven't fixed this because i thought it would be nice if the extrema changed when you hover over a graph, like the current value does. Thoughts?

EDIT - Just realised I also left in the ability to turn off compressing the localForage storage. I think it's handy for debugging, but you might not want it cluttering the code.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -30,7 +30,7 @@ localForage.config({
 });
 
 localForage.iterate((data, key) => {
-  const value = decompress(data);
+  const value = key.endsWith('-raw') ? data : decompress(data);
   const start = new Date();
   start.setHours(start.getHours() - value.hours_to_show);
   if (new Date(value.last_fetched) < start) {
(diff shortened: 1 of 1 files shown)


kalkih:
Thanks @michaelblight!

I'm currently on vacation so I just had a quick peek, looks great!

As for hiding the first entity, yeah it's an issue, but it would require quite a bit of additional logic to fix it (I think), but yes, would be nice to fix eventually.

Quick thought. Calculating the extrema for other entities would be cool, but requires more computing, doing it on hover could potentially be too demanding? and make the card lag for a short sec (at least on lower perf devices), we could of course do it prior and save the results, what do you think?

Turning of the compression is useful, let's leave it in there.

Thanks again @michaelblight, great work!

Issue #73, opened 2019-03-14
Request: Option for showing average


klaasnicolaas:
I noticed that when I put 2 temperature sensors in a graph, the min and max are still calculated from only the first entity. The 11.7 degrees also comes from 1 entity.

![Sensor](https://user-images.githubusercontent.com/20448157/54388230-94193300-469d-11e9-97d7-a967810ac78c.png)

Could something be made that grabs an average? Like now I have 2 sensors in the same room, at 2 different places in the room (hence the differences). Only the 11.7 degrees is then just not a representative temperature to show.

kalkih:
Yes, the extrema view does currently only display information from the primary entity (first one listed in the config).

Support for displaying the extremum per entity basis has been suggested here: #51.
I don't think an *average extrema* will be implemented, kinda defeats the purpose of having an extremum.

For your use case I would recommend creating a template sensor that stores the average state of your two sensors, and then use that template sensor in the `mini-graph-card`.

kalkih:
We now have support for the average Thanks to @TheLastProject #135.
If you want the average of two sensors I would recommend creating a template sensor.

PR #135, opened 2019-08-16
Add average support

TheLastProject:
Note: not related to #73, this shows the average of the first value only, just like the current extrema support. It may however be helpful to that reporter anyway if they did create an extra sensor.

```
show:
  extrema: true
  average: true
```
Screenshot 2019-08-16 at 15 25 08

```
show:
  average: true
  extrema: false
```
Screenshot 2019-08-16 at 15 24 29

```
show:
  extrema: true
  average: false
```
Screenshot 2019-08-16 at 15 23 37


kalkih:
Nice, thanks for the PR!

Issue #31, opened 2019-01-31
Enable popup support for multiple entities


kalkih:
Pressing/clicking legend entries will trigger the popup for that entity

kalkih:
Now available by pressing the legend items in [v0.2.1](https://github.com/kalkih/mini-graph-card/releases/tag/v0.2.1).

PR #136, opened 2019-08-19
Pressing/clicking state entries will trigger the popup for that entity

jbalague:
This intends to complement issue #31, which does the same when clicking legend entries.


kalkih:
Awesome, great addition, thanks!

PR #149, opened 2019-09-06
Adding more colors for automatic multi colored graph

### LINKED ISSUE #147: Feature Request: More default colors
Hey,

Love this card, and using it for graphs with more that 6 entities. But there's only 6 default colors. What about using more ?
Using a palette like this one : http://colrd.com/palette/27645/ with 16 colors would be nice no ?

If you think that's nice, i'll be glad to PR this.

SNoof85:
With this, there would be 13 colors for multi entities graphs.

Preview : 
image

Note that `var(--accent-color)` is not in the preview !
I preserved the original order for firsts colors to avoid any change in small multiple entities existing graphs.

Colors from https://flatuicolors.com/palette/defo

Would close #147

### DIFF
--- a/src/const.js
+++ b/src/const.js
@@ -21,7 +21,7 @@ const DEFAULT_COLORS = [
   '#7f8c8d',
   '#27ae60',
   '#2980b9',
-  '#8e44ad'
+  '#8e44ad',
 ];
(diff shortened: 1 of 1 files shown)


kalkih:
Perfect, great addition.
Thank you very much!

Issue #78, opened 2019-04-02
Daily bar chart


tinglis1:
Feature request:
It would be great to be able to make a daily bar chart.
I can currently set a 24hr bar chart but it would be great if the period could be locked to a set period e.g. midnight to midnight.

My use case for this would be to show energy usage for the last 7 days from a cumulative energy sensor. The current day would only show the usage from midnight.

kalkih:
Yes indeed, that would be a great and useful feature.

mrkrb:
I agree as well, this would be awesome! For example for creating charts of the min and max temperature along the week, or to display the power consumption of the last N months, so on so forth.

NicoBijl:
Would be very useful indeed, for me personally to visualize my hours at work on a daily basis (based on device tracking)

mynameisdaniel32:
Something that could fit well with this would be an option for a pie chart - who doesn't need more pie charts in their life?

arnoldpistorius:
My use case would be to view the distance traveled with my car per day in a bar chart.

kaosmagix:
+1 For displaying my daily utility totals (they reset at midnight)

kaosmagix:
see also issue #77

geniusve:
+1 For displaying my daily energy consumption.

maxwroc:
I needed it too. Started working on the change today and it looks like I have a working code already. Once I test it and make it prod ready will send PR (probably over the weekend)

maxwroc:
PR sent and waiting for review

ACrazyConcept:
Just tried the new bar chart and I can't get it working correctly:
`      - type: custom:mini-graph-card
        entities:
          - entity: sensor.daily_electricity
        name: Energy consumption
        hours_to_show: 168
        aggregate_func: max
        group_by: date
        show:
          graph: bar
`

maxwroc:
It looks like the change is not in your js. Please make sure that your browser doesn't use cached previous version. Do the following:
* in your ui-lovelace.yaml add version ("v") param int he url e.g. resources:
  url: /community_plugin/mini-graph-card/mini-graph-card-bundle.js?**v=0.7.1**
* refresh lovelace UI by clicking on ".." up-right corner of your Home Assistant and then click "Refresh"
* force your browser to download fresh resources by using Ctrl+F5 combination

Please let me know if that helps.

kalkih:
[...]

Issue #51, opened 2019-02-14
Extrema View - improvments


SeLLeRoNe:
It would be nice on the Extrema View to be able to see one line per entity so to keep track of all min-max of each entity of the graph.

It would also be nice, when there are multiple entities but you want to see only one min/max, to be able to define which entity provides you the min and which the max.

Hope is clear enough, if not let me know :)

Cheers
Andrea

kalkih:
I think we could add this as an optional option in the [entity object](https://github.com/kalkih/mini-graph-card#entities-object), similar to `show_state`.

Possibly something like this:
```yaml
entities:
  - sensor.temperature
  - entity: sensor.pressure
    name: Pressure
    show_state: true
    show_extrema: true #show this entity in the extrema
  - sensor.humidity
```

SeLLeRoNe:
Yep I think that would make a lot of sense.

It might actually also work in another way:
```
extrema: true
entities:
  - sensor.temperature
  - entity: sensor.pressure
    name: Pressure
    show_state: true
    show_extrema: false #don't show this entity in the extrema but show all the rest
  - sensor.humidity
```

ahd71:
When displaying data over more than 24h it shows only TIME of the extreme value, would be good to be able to format the value to show DATE+TIME (in 12/24h, and with different date formats, i.e yyyy-mm-dd, preferably from some system setting)

kalkih:
> When displaying data over more than 24h it shows only TIME of the extreme value, would be good to be able to format the value to show DATE+TIME (in 12/24h, and with different date formats, i.e yyyy-mm-dd, preferably from some system setting)

The day / weekday will be displayed by default in the next release if the graph timeframe is longer than 24 hours d1bdfce64a285e9bf9ccb068a668ebef021cb954

The format will be (D ddd, HH:mm - HH:mm ) or corresponding am/pm format

ahd71:
Excellent, thanks!

hekm77:
Is it possible to display the weekdays in a local HA language?

kalkih:
@hekm77 

Sure, coming next update https://github.com/kalkih/mini-graph-card/commit/6d542863695e62a3d44dc089cc8ab82f3472a085

kalkih:
@hekm77 This is now available in the latest release [v0.3.2](https://github.com/kalkih/mini-graph-card/releases/tag/v0.3.2), no config needed, utalize the language set in HA.

mihalski:
[...]

PR #165, opened 2019-10-11
Group by date and aggregate functions #78 #51

maxwroc:
Daily results and support for various aggregation methods.

Main goal of this change was to give ability to show utility meter results. It other words show grouped results by day and the maximum value achieved - which means show results from 00:00 till 23:59 as a single point.

I've added additional aggregation methods (other than maximum) and now we can for example show min, max, avg temperature stats on the same graph.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,7 @@ properties of the Entity object detailed in the following table (as per `sensor.
 | name | string |  | Set a custom display name, defaults to entity's friendly_name.
 | color | string |  | Set a custom color, overrides all other color options including thresholds.
 | unit | string |  | Set a custom unit of measurement, overrides `unit` set in base config.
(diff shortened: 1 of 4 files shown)


maxwroc:
#78 #51 Both issue could be solved here

kalkih:
Fantastic, two great additions!
Adds a bunch of new use cases for the card.
Great execution on the implementation.

Regarding the liner, it's actually complaining about the reassignment of the func parameter [no-param-reassign](https://eslint.org/docs/rules/no-param-reassign).
Strange thing is, you can kind of side step the rule by naming the func param `res`, any other name (I've tried a couple for science) will result in a linter error. I found that pretty interesting.

Thank you for your work on this PR, nothing to complain about 👍

Issue #63, opened 2019-02-22
request: a way to overlay binary sensor data


skarfacegc:
One of my use cases for graphing is to show my hvac state.  I'm currently using this to show set temp, actual temp, and when the hvac system is heating or cooling  (the bar across the top).  It would be super useful to be able to include the binary sensor data in the graph (perhaps as an area fill under the primary graph value, or even just a line of status colors across the bottom)  

I've tried turning the heat/cool/off values into 1/0 and putting those as a line across the bottom but the graph smoothing (I think) gets in the way of the line being hidden when the value is 0.

<img width="391" alt="screen shot 2019-02-22 at 11 41 30 am" src="https://user-images.githubusercontent.com/1668735/53256869-e5ce2f00-3696-11e9-8cb1-e230e8632b8b.png">

kalkih:
You could always put a `history-graph` with just the binary sensor just below/above the mini graph card. Possibly wrap them both in a `vertical-style-card`.

skarfacegc:
yeah, that's what I have now.  Thought it would be cool to have the state history as an overlay on the main graph.

kalkih:
Okay. Most likely not going to add this though, I want to keep this card simple and minimalistic, the `history-graph` is a better alternative for more advanced or detailed use cases.

francesconazzaro:
It would be great something like this one. I didn't find yet some way to have a graph like this in home assistant.

![image](https://user-images.githubusercontent.com/9554021/59035558-d356ba80-886d-11e9-951c-d2b55fa4c8a4.png)

dlashua:
I came here looking for the same thing... a way to overlay the state of a binary_sensor.

Interesting, too, because the README shows the following config:
```
- type: custom:mini-graph-card
  entities:
    - entity: sensor.verandah
      name: Verandah
    - entity: sensor.lounge
      name: Lounge
    - entity: sensor.kitchen
      name: Kitchen
    - color: gray
      entity: input_number.nighttime
      name: Night
      show_line: false
      show_points: false
      show_legend: false
      y_axis: secondary
    show:
      labels: true
      labels_secondary: true
```

... where, looking at the graph, input_number.nighttime is either 1 or 0 and the graph displays exactly in the way I'm looking for.

As a work around, a template sensor like this will work just as well. However, you'll have to wait a few days for the data to populate to get anything useful to look at.

```
sensor:
  - platform: template
    sensors:
      numeric_binary_sensor:
[...]

Issue #117, opened 2019-07-10
Graph not displaying correct values. 


tomlut:
I'm seeing strange behaviour of the graph. It is reporting a high value when it shouldn't. See screen caps below:

![Annotation 2019-07-10 104738](https://user-images.githubusercontent.com/10679300/60932540-c5c7a100-a301-11e9-80f4-b88d0d3287a8.png)
![Annotation 2019-07-10 104822](https://user-images.githubusercontent.com/10679300/60932542-c5c7a100-a301-11e9-994e-f6c54b0d1256.png)

No amount of changing the points per hour seems to fix it.

```
      - color_thresholds:
          - color: '#0da035'
            value: -1
          - color: '#e0b400'
            value: 1800
          - color: '#e45e65'
            value: 2100
        color_thresholds_transition: hard
        entities:
          - sensor.lounge_dehumidifier_power
        group: false
        hour24: true
        line_width: 6
        name: Lounge Dehumidifier Power
        points_per_hour: 2
        show:
          fill: fade
          icon: true
          labels: false
          name: true
          state: true
        style: |
          ha-card {
            border: solid 2px var(--primary-color);
          }
        type: 'custom:mini-graph-card'
```

<s>EDIT: this appears to be a firefox issue. The graphs display correctly on Chrome and Safari. Clearing the cache on firefox (CTRL + F5) did not make any improvement.</s>

EDIT2: or maybe it is a caching issue. Opening the page in a private window in firefox does not have the problem.

kalkih:
Could it be that the sensor has no history for that initial timespan? If there's no history data available at the beginning of the timespan we don't know how to fill the graph, and will give the "missing timespan" the value of the first known history entry, which in this case looks to be about 16? hours in.

The default history graph I believe does this differently and will fill the missing timespan with the value of the previous known history entry in existence, even if it's prior to the actual timeframe of the graph.

I would guess this is the reason to why you are seeing this behaviour, but could also be something else, hard to tell without the actual history data.

It's strange though, if it's behaving differently if cached or not, what did the graph look like when you opened it in incognito or another browser? We might have an inconsistency in how we compute the history data when it's cached versus fetched.

tomlut:
My claim that Chrome was ok is false. It is happening there too now.

Looks like this in a previously opened browser:

![previously_open](https://user-images.githubusercontent.com/10679300/60968897-892b9200-a361-11e9-811d-8665fb91b28e.png)

Looks like this in a private tab:

[...]

PR #170, opened 2019-10-15
Support for non numeric states (+ disable smoothing setting) - #63

maxwroc:
### Feature
With this change we will be able to show e.g. binary sensor states on the graph. This will fill a big gap as the default way of rendering of such sensor data is not readable (talking here about timeline horizontal bars rendered by history-graph card (or in more-info dialog)).

Feature requested in #63 

### Bug fix
In addition I have fixed here the initial state which comes from cache. The problem was that we were filtering out older states than current time but we were not adding previous state on the beginning of the time. This way when sensor was pushing data rarely we were rendering graph in the wrong way.
For example: 
* time now: 2019-10-15 **14:30:00**
* graph beginning time: 2019-10-15 **13:30:00**
* data in cache: 
  ```js
  [ 
    {last_changed: '2019-10-15 12:10:00', state: 25 },
    {last_changed: '2019-10-15 13:50:00', state: 30 },
    {last_changed: '2019-10-15 14:25:00', state: 28 }
  ]
  ```
* filtered result: 
  ```js
  [ 
    {last_changed: '2019-10-15 13:50:00', state: 30 },
    {last_changed: '2019-10-15 14:25:00', state: 28 }
  ]
  ```
So now we will show straight line from the beginning of the graph (13:30) with value 30 till 13:50:00 which is not correct. The line should start on value 25 and it should change at 13:50 to 30.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -743,6 +743,9 @@ class MiniGraphCard extends LitElement {
 
       if (stateMap) {
         return stateMap.label;
+      } else {
+        // eslint-disable-next-line no-console
+        console.warn(`mini-graph-card: value [${inState}] not found in state_map`);
       }
     }
 
--- a/src/main.js
+++ b/src/main.js
@@ -737,9 +737,10 @@ class MiniGraphCard extends LitElement {
 
   computeState(inState) {
     if (this.config.state_map.length > 0) {
-      const stateMap = Number.isInteger(inState) ?
-                        this.config.state_map[inState] :
-                        this.config.state_map.find(state => state.value === inState);
+      const stateMap = Number.isInteger(inState)
+        ? this.config.state_map[inState]
+        : this.config.state_map.find(state => state.value === inState);
+
       if (stateMap) {
         return stateMap.label;
       }
@@ -842,17 +843,16 @@ class MiniGraphCard extends LitElement {
     if (history && history.hours_to_show === this.config.hours_to_show) {
       stateHistory = history.data;
 
-      let currDataStartIndex = stateHistory.findIndex(item => new Date(item.last_changed) > initStart);
-      if (currDataStartIndex !== -1) {
-
-        if (currDataStartIndex > 0) {
+      let currDataIndex = stateHistory.findIndex(item => new Date(item.last_changed) > initStart);
+      if (currDataIndex !== -1) {
+        if (currDataIndex > 0) {
           // include previous item
-          currDataStartIndex--;
+          currDataIndex -= 1;
           // but change it's last changed time
-          stateHistory[currDataStartIndex].last_changed = initStart;
+          stateHistory[currDataIndex].last_changed = initStart;
         }
 
-        stateHistory = stateHistory.slice(currDataStartIndex, stateHistory.length);
+        stateHistory = stateHistory.slice(currDataIndex, stateHistory.length);
         // skip initial state when fetching recent/not-cached data
         skipInitialState = true;
       }
@@ -925,16 +925,13 @@ class MiniGraphCard extends LitElement {
   }
 
 
-  _convertState(item) {
-    const resultIndex = this.config.state_map.findIndex(stateData => stateData.value === item.state);
+  _convertState(res) {
+    const resultIndex = this.config.state_map.findIndex(s => s.value === res.state);
     if (resultIndex === -1) {
-      console.warn(`mini-graph-card: entity state cannot be converted: ${item.state}`);
       return;
     }
 
-    item.state = resultIndex;
-
-    return;
+    res.state = resultIndex;
   }
 
   getCardSize() {
--- a/src/main.js
+++ b/src/main.js
@@ -138,7 +138,8 @@ class MiniGraphCard extends LitElement {
       color_thresholds_transition: 'smooth',
       line_width: 5,
       compress: true,
-      state_map: null,
+      smoothing: true,
+      state_map: [],
       tap_action: {
         action: 'more-info',
       },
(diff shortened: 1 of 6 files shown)


src/main.js
```diff
@@ -725,6 +736,16 @@ class MiniGraphCard extends LitElement {
...
 
   computeState(inState) {
+    if (this.config.state_map.length > 0) {
+      const stateMap = Number.isInteger(inState)
+        ? this.config.state_map[inState]
+        : this.config.state_map.find(state => state.value === inState);
+
+      if (stateMap) {
+        return stateMap.label;
+      }
```
maxwroc: Wanted to console log/warn something to tell user that theirs configuration is not complete but eslint complains

src/main.js
```diff
@@ -821,10 +842,21 @@ class MiniGraphCard extends LitElement {
...
-      if (stateHistory.length > 0) {
+
+      let currDataIndex = stateHistory.findIndex(item => new Date(item.last_changed) > initStart);
+      if (currDataIndex !== -1) {
+        if (currDataIndex > 0) {
+          // include previous item
+          currDataIndex -= 1;
+          // but change it's last changed time
+          stateHistory[currDataIndex].last_changed = initStart;
+        }
```
maxwroc: That is the actual bugfix described in the PR

src/main.js
```diff
@@ -887,6 +924,16 @@ class MiniGraphCard extends LitElement {
     return this._hass.callApi('GET', url);
   }
 
+
+  _convertState(res) {
```
maxwroc: Again used this "res" name trick to make eslint happy but actually in this case I could just return the item/res and use assignment and map function above - where I call _convertState. Up to you - please let me know :)

src/main.js
kalkih: Yep, that's a good idea! You could add an exception for the line, like below:

https://github.com/kalkih/mini-graph-card/blob/34142fdefa5c1a4292c67acb3eda4ff1475fe9b6/src/main.js#L890-L891

src/main.js
kalkih: Very clever!
That should solve https://github.com/kalkih/mini-graph-card/issues/117 as well I believe?

src/main.js
kalkih: I don't mind either really, as long as eslint is happy lol, and we do it on a few other placed already so I wouldn't worry about it. 😉

src/main.js
maxwroc: yes it looks like it is the same root cause

src/main.js
maxwroc: Ok will add it

kalkih:
Once again, great additions @maxwroc!

I noticed that points still get assigned the "smoothed" value.

Kapture 2019-10-15 at 15 12 36

So we prob need to change this line as well.
https://github.com/kalkih/mini-graph-card/blob/34142fdefa5c1a4292c67acb3eda4ff1475fe9b6/src/graph.js#L113

Possibly:
```js
const sum = this._smoothing ? (next[V] + last[V]) / 2 : next[V];
```

kalkih:
Adding a normal sensor to the graph (and same y-axis) will kind of break this.

hyperion_8123_lovelace_example (16)

Not sure what we would do about this though.


Also, do you think we should force `smoothing: false` or possibly set as default for `binary_sensor` entities? I can't think of a use case where you'd want it on a binary_sensor.

src/main.js
kalkih: We could/should probably do a very simple log helper function in the future.

maxwroc:
> ```js
> this._smoothing ? (next[V] + last[V]) / 2 : next[V];
> ```

Nice catch! I've missed it.

maxwroc:
> Once again, great additions @maxwroc!
> 
> I noticed that points still get assigned the "smoothed" value.
> 
> Kapture 2019-10-15 at 15 12 36
> 
> So we prob need to change this line as well.
> https://github.com/kalkih/mini-graph-card/blob/34142fdefa5c1a4292c67acb3eda4ff1475fe9b6/src/graph.js#L113
> 
> Possibly:
> 
> ```js
> const sum = this._smoothing ? (next[V] + last[V]) / 2 : next[V];
> ```

Nice catch! I have missed it

maxwroc:
> Adding a normal sensor to the graph (and same y-axis) will kind of break this.
> 
> hyperion_8123_lovelace_example (16)
> 
> Not sure what we would do about this though.
> 
> Also, do you think we should force `smoothing: false` or possibly set as default for `binary_sensor` entities? I can't think of a use case where you'd want it on a binary_sensor.

I wouldn't turn on smoothing by default for all the graphs. Turning it on for binary sensors is an interesting idea and I think it makes sense (but I'd like to leave the option for user to override it). Though it won't be so straightforward as on the time when `setConfig` is called I don't have the info about entity type yet (`hass` hasn't been called at this point yet). 
Will try to figure out something. Thanks for suggestion.

Regarding y-axis issue: these are completely different sensors with different values so in such case the `y_axis: secondary` option should be used in the card config. Then everything is fine, right?

maxwroc:
OK I have used entity name to figure out the type (don't know why I wanted to use entity_id from states as it has exactly the same val)

kalkih:
> Regarding y-axis issue: these are completely different sensors with different values so in such case the `y_axis: secondary` option should be used in the card config. Then everything is fine, right?

Indeed, putting them on different axes solves the issue, just not optimal usability wise, adds another step for the user, but It's fine, at least for now imo. 🙂

maxwroc:
I have just logged a potential issue #171 

I didn't have time to test it and I'm not sure if I find some time soon, sorry about that. Wanted to create the item to track this potential bug. Please close it if this is not the actual case.

PR #172, opened 2019-10-17
Support for group by hour

kalkih:
Adds support for grouping data by hour, similar to group by date #165

### DIFF
--- a/src/graph.js
+++ b/src/graph.js
@@ -8,6 +8,11 @@ export default class Graph {
       max: this._maximum,
       min: this._minimum,
     };
+    const groupByFuncMap = {
+      interval: this._getGroupByIntervalFunc(),
+      date: this._getGroupByDateFunc(),
+      hour: this._getGroupByHourFunc(),
+    };
 
     this.coords = [];
     this.width = width - margin[X] * 2;
@@ -17,8 +22,8 @@ export default class Graph {
     this._min = 0;
     this.points = points;
     this.hours = hours;
-    this._calculatePoint = aggregateFuncMap[aggregateFuncName] || this._average;
-    this._groupBy = groupBy;
+    this._calcPoint = aggregateFuncMap[aggregateFuncName] || this._average;
+    this._reducer = groupByFuncMap[groupBy] || this._getGroupByIntervalFunc;
     this._smoothing = smoothing;
   }
 
@@ -31,8 +36,7 @@ export default class Graph {
   set min(min) { this._min = min; }
 
   update(history) {
-    const groupByFunc = this._groupBy === 'date' ? this._getGroupByDateFunc() : this._getGroupByIntervalFunc();
-    const coords = history.reduce((res, item) => groupByFunc(res, item), []);
+    const coords = history.reduce((res, item) => this._reducer(res, item), []);
 
     const requiredNumOfPoints = Math.ceil(this.hours * this.points);
     if (coords.length > requiredNumOfPoints) {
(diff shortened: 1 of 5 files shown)


src/graph.js
```diff
@@ -72,17 +76,31 @@ export default class Graph {
     };
   }
 
+  _getGroupByHourFunc() {
+    const now = new Date();
+    now.setHours(now.getHours() + 1);
+    now.setMinutes(0, 0, 0);
```
maxwroc: I have noticed now an issue. This code is executed once when graph is being initialized. This means that "now" won't update when the `update` function is triggered. We need to fix it as well in interval grouping implementation

src/graph.js
```diff
@@ -72,17 +76,31 @@ export default class Graph {
...
+    return (res, item) => {
+      const age = now - new Date(item.last_changed).getTime();
+      const interval = (age / (1000 * 3600)) - this.hours;
```
maxwroc: nitpick: I saw it somewhere else too - it would be good to have a const for it. It will improve code readability too (talking about `1000 * 3600`)

src/main.js
```diff
@@ -574,6 +576,9 @@ class MiniGraphCard extends LitElement {
       // move end time to the next day at midnight
       now.setDate(now.getDate() + 1);
       now.setHours(0, 0);
+    } else if (group_by === 'hour') {
+      now.setHours(now.getHours() + 1);
+      now.setMinutes(0, 0);
```
maxwroc: same here

src/main.js
```diff
@@ -170,9 +170,11 @@ class MiniGraphCard extends LitElement {
     const additional = conf.hours_to_show > 24 ? { day: 'numeric', weekday: 'short' } : {};
     conf.format = { hour12: !conf.hour24, ...additional };
 
+    // override points per hour to mach group_by function
     if (conf.group_by === 'date') {
-      // override points per hour to mach 1 point per day (24h)
       conf.points_per_hour = 1 / 24;
+    } else if (conf.group_by === 'hour') {
+      conf.points_per_hour = 1;
```
maxwroc: consider switch instead - easier to extend in the future and better readability IMO

src/main.js
```diff
@@ -574,6 +576,9 @@ class MiniGraphCard extends LitElement {
       // move end time to the next day at midnight
       now.setDate(now.getDate() + 1);
       now.setHours(0, 0);
+    } else if (group_by === 'hour') {
+      now.setHours(now.getHours() + 1);
+      now.setMinutes(0, 0);
     }
```
maxwroc: Was thinking about this too. Thanks for adding it

maxwroc:
Nice! I was thinking about it too. Thanks for adding it

src/graph.js
kalkih: I actually thought about this when I implemented the new hour function.

What do you think about moving it to a private variable and update it when the update function is called?

src/graph.js
kalkih: Sure!

kalkih:
Thanks for your feedback @maxwroc, will address tmrw.
I added you as a collaborator to the repo if you wanna work/push directly to it etc... 👍

maxwroc:
Pushing directly is not so exciting, i can still do it from my fork and that's not a problem but if I could get notification about new PRs then I could help you with them. I like this project and I'll keep an eye on it for sure either way.

kalkih:
Thanks, I appreciate it.
You should be able to interact with PRs as a collaborator, e.g. label/review/merge.

maxwroc:
Pushing directly is not so exciting, i can still do it from my fork and that's not a problem but if I could get notification about new PRs that would be cool. I could help you with them. I like this project and I'll keep an eye on it for sure either way.

src/graph.js
maxwroc: Depends what do you want to move. Is it “new now date” or sth else. I would go for putting group_by as a private prop and reassign the func on the update event - it should be cheap anyway and this way you will still have all required vars in the same local scope

maxwroc: you need to generate new Date on every update anyway

maxwroc:
Cool thx.

src/graph.js
kalkih: Think I found a good solution, this way I could get rid of the new `_getGroupByHourFunc()` as well and reuse `_getGroupByIntervalFunc()`

src/main.js
kalkih: Replaced

Issue #183, opened 2019-10-22
group_by: date not working as expected


stepanov1975:
I am using following definition
```
- type: custom:mini-graph-card
  entities:
    - entity: sensor.socket_dishwasher_energy_today
  name: Dishwasher
  aggregate_func: max
  group_by: date
  hours_to_show: 168
  show:
    graph: bar
```
So I expect that for each day maximum value calculated and single bar for this day will be present on the graph.
Instead, I see the following:
![Capture1](https://user-images.githubusercontent.com/9785479/67312514-202fea80-f50a-11e9-9924-93e40be4fb1b.PNG)
What am I doing wrong?

kalkih:
Hey,
Looks like your browser isn't loading the latest version of the card.
If you've installed the card through HACS please check this: https://github.com/kalkih/mini-graph-card/issues/169
If you installed it manually, try clearing cache.

stepanov1975:
Thanks. I fixed path and graph looks like expected, but values are wrong
Graph:
![Capture1](https://user-images.githubusercontent.com/9785479/67314524-75213000-f50d-11e9-83be-d885422f47cd.PNG)
Actual values:
![Capture](https://user-images.githubusercontent.com/9785479/67314608-a26dde00-f50d-11e9-855a-055917a4c786.PNG)

kalkih:
Yeah that looks weird, does both graph use the same data source, the HA history database?

What does it look like if you do a line graph with the same data:

```yaml
- type: custom:mini-graph-card
  entities:
    - entity: sensor.socket_dishwasher_energy_today
  name: Dishwasher
  hours_to_show: 168
```

Any ideas @maxwroc ?

stepanov1975:
No, the second one from InfluxDB but the date should be same
And the line graph looks right
![Capture](https://user-images.githubusercontent.com/9785479/67316136-93d4f600-f510-11e9-92d2-d8f5f5e6d4c9.PNG)

maxwroc:
I don't have time to look in the code now but i suspect where the problem might be located...

There is this place where we set the result array length. It is little bit odd anyway. And I think that what happens here is that there is not enough data to cover 7 days, we put the available data on the beginning of the array and then extend it to 7 items. 
The result is that first bars show correct values and the rest just repeat the last value.
This is corner case which I haven't tested. To be honest I don't know then how come it works for line graphs :) as the logic is the same.
If you have time before weekend you can try to fix it if not I'll do it on Fri/Sat

PR #187, opened 2019-10-23
Group by date missing data

kalkih:
This will add missing array entries at the beginning instead of at the end when using `group_by: date`, it's  not a great solution though since it doesn't cover all cases. With this fix we basically anticipate; if length is incorrect we are missing data at the beginning.

So I think if we are missing data at the end or for a whole date when using `group_by: date` the graph will still render incorrectly

The `_getGroupByIntervalFunc()` works by inserting data at the correct index in the coords array, so for example if we have:

* hours_to_show: 6
* points_per_hour: 1

and we are missing data for hour 1,3 and 6, the output would be:

`[empty, data, empty, data, data]`

Therefore if the array is ever too small, we know we are missing data at the end (most recent) of the timeframe and we just increase the length of the array to fill the missing data with empty entries.

Length is less than expected (hours * points), expand length:

`[empty, data, empty, data, data, empty]`

Fixes: #183

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -577,7 +577,6 @@ class MiniGraphCard extends LitElement {
       points_per_hour,
       hours_to_show,
       format,
-      group_by,
     } = this.config;
     const offset = hours_to_show < 1 && points_per_hour < 1
       ? points_per_hour * hours_to_show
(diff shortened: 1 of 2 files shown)


kalkih:
@maxwroc

kalkih:
I think that should do it!
the group by date function now work similar to the group by interval function.

I'm not sure if we need this anymore?
https://github.com/kalkih/mini-graph-card/blob/318dc473ea0c16075c8421037d921480e3d385a1/src/graph.js#L48-L50

kalkih:
Last commit will make sure we don't drop the initial state in `group_by` `hour` and `date`, and will also prevent us from fetching unnecessary data before our rendered timeframe (`group_by` `hour` or `date`) that we would otherwise just throw away.

maxwroc:
Looks really nice! Testing...

maxwroc:
> ...and will also prevent us from fetching unnecessary data before our rendered timeframe (`group_by` `hour` or `date`) that we would otherwise just throw away.

To be honest I don't see that part in modified code. Fetching logic is on main.js right?

kalkih:
Haha whoops, you would have if I hadn't closed my terminal before the push finished 😆

src/main.js
```diff
@@ -802,7 +788,7 @@ class MiniGraphCard extends LitElement {
   async updateData({ config } = this) {
     this.updating = true;
 
-    const end = new Date();
+    const end = this.getEndDate();
```
kalkih: This is what I was referring to 😉

maxwroc: :)

PR #188, opened 2019-10-25
Bar Click Event for Desktop, Numberformat

zibous:
Add barclick event for desktop browers.

Format Number based on locale

```
## new settings property
locales: 'de-DE'
```
Sample:

numberformat

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -140,6 +140,7 @@ class MiniGraphCard extends LitElement {
       color_thresholds: [],
       color_thresholds_transition: 'smooth',
       line_width: 5,
+      locales: null,
       compress: true,
       smoothing: true,
       state_map: [],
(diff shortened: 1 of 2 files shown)


kalkih:
I like the locale option, please add it to the readme.

src/main.js
```diff
@@ -777,10 +787,19 @@ class MiniGraphCard extends LitElement {
...
+  }
+
+  formatStateValue(v, dec) {
+    const _locales = this.config.locales;
+    if (_locales === null || _locales === undefined) {
+      return v;
+    } else {
+      return parseFloat(v, 10).toLocaleString(_locales, { maximumFractionDigits: Number(dec) });
+    }
   }
```
kalkih: Could replace with
```js
formatStateValue(state, maximumFractionDigits) {
  if (this.config.locale) {
    return parseFloat(state, 10).toLocaleString(this.config.locale, { maximumFractionDigits });
  }
  return state;
}
```

Not sure if we need the parseFloat() since we always do Number() or parseFloat() in computeState() before calling this function.

PR #235, opened 2020-01-11
fix missing entities

### LINKED ISSUE #234: Ability to ignore empty entities (and maybe in the future also invalid ones)
Hey there,
I was wondering if there is the possibility to ignore invalid/empty entities in the config

```
      entities:
        - entity: '[[entity]]'
          name: Temperature
          color: '#FF5000'
          show_fill: false
        - entity: '[[entity2]]'
          name: Outside
          color: '#0099CC'
          show_state: true
          show_fill: true
        - entity: input_number.night_time
          color: '#B3B3B3'
          name: Night Time
          show_fill: true
          show_line: false
          show_points: false
          show_legend: false
          y_axis: secondary
```

I am using this in combination with the decluttering card and I can send variables ([[entity]]) this way:
```
              - type: custom:decluttering-card
                template: temperature
                variables:
                  - entity: sensor.patio_temperature
```

In the even (or requirement) that [[entity2]] comes empty, would be possible to simply not showing the specific entity rather then having the card fail?

```
Uncaught (in promise) TypeError: Cannot destructure property 'state' of 'this.entity[b]' as it is undefined.
    at HTMLElement.renderState (mini-graph-card-bundle.js:1)
    at mini-graph-card-bundle.js:1
    at Array.map (<anonymous>)
    at HTMLElement.renderStates (mini-graph-card-bundle.js:1)
    at HTMLElement.render (mini-graph-card-bundle.js:1)
    at HTMLElement.update (mini-graph-card-bundle.js:1)
    at HTMLElement.performUpdate (mini-graph-card-bundle.js:1)
    at HTMLElement._enqueueUpdate (mini-graph-card-bundle.js:1)
```
```
Uncaught (in promise) TypeError: Cannot destructure property 'state' of 'this.entity[b]' as it is undefined.
    at HTMLElement.renderState (mini-graph-card-bundle.js:1)
    at mini-graph-card-bundle.js:1
    at Array.map (<anonymous>)
    at HTMLElement.renderStates (mini-graph-card-bundle.js:1)
    at HTMLElement.render (mini-graph-card-bundle.js:1)
    at HTMLElement.update (mini-graph-card-bundle.js:1)
    at HTMLElement.performUpdate (mini-graph-card-bundle.js:1)
    at HTMLElement._enqueueUpdate (mini-graph-card-bundle.js:1)
```

Thanks

nurikk:
Ability to ignore empty entities.
Fixes #234 
Missing entity causes crash for plugin.
`mini-graph-card-bundle.js:1 Uncaught (in promise) TypeError: Cannot read property 'entity_id' of undefined`

This can happen if your sensor is gone for a while example: your configured chart with 3 temp sensors
living room/bedroom/kitchen. But suddenly one of sensors is gone (battery discharged). 

`      
- entities:
          - entity: sensor.mi_t_4c65a8d8de16
            name: Living
          - entity: sensor.xiaomi_miio_device_temperature
            name: Bedroom
          - entity: sensor.temperature_158d0001fa4f2f
            name: Dining
        name: Temperature
        type: "custom:mini-graph-card"
`


kalkih:
Hello, thanks for the PR!

This should work great when entities are missing/undefined during Initialization of the card but I think we should remove additional saved entity data if we remove an entity later during runtime.
The reason for this being that we use the position of the entity in the config to determine what and how to render the entity, this will get mapped incorrectly if we remove the entity from our config array and keep it in our other data stores such as line, point & bar data e.t.c.

I think something like this should solve the issue? Could potentially move the ugly array to the const file.
```js
this.config.entities = this.config.entities.filter((entry, i) => {
  console.warn(`No state data for entity ${entry.entity}, removing it from config`);
  if (!hass.states[entry.entity]) {
    ['length',
      'entity',
      'line',
      'bar',
      'fill',
      'points',
    ].forEach((attr => this[attr].splice(i, 1)));
    return false;
  } else {
    return true;
  }
});
```

Another issue I see is that we have no way of adding entities back if they reappear in HA again, not sure how important this is or if we should leave it.

What do you think?

nurikk:
Yes, your are right! Removing entities from config in runtime was bad idea. I refactored my implementation, now its flags suitable entities for rendering and better handles default values

src/main.js
```diff
@@ -684,7 +688,7 @@ class MiniGraphCard extends LitElement {
   }
 
   get visibleEntities() {
-    return this.config.entities.filter(entity => entity.show_graph !== false);
+    return this.config.entities.filter(entity => entity.enabeled && entity.show_graph !== false);
   }
```
kalkih: We can't really do it like this to decide what to render throughout the card since `show_graph` should only hide the entity from the "graph" part of the card, the state would still be visible e.t.c.

kalkih:
Btw, do we want to remove an entity from the graph if it goes missing? I guess we don't really need to?

src/main.js
nurikk: why we can't? to be visible graph should be valid(enabled) and show_graph, this logic completely solves current problem

nurikk:
why not? as soon as it come back we will draw it again,  since set hass(hass) is called for every state update

src/main.js
kalkih: `show_graph: false` does not mean the user want to hide the entity from the card completely, it means the entity should not render in the graph but a user might still want to see the state, extrema e.t.c.

So if we use `visibleEntities` to decide what state to render we won't render the state of an entity that has `show_graph: false` set even though we should. 
https://github.com/nurikk/mini-graph-card/blob/fix_missing_entity/src/main.js#L328-L329

kalkih:
> why not? as soon as it come back we will draw it again, since set hass(hass) is called for every state update

Your current PR will still render missing entities in the graph, we would have to clear the position of the entity in all the following arrays ('length', 'entity', 'line', 'bar', 'fill', 'points') since they aren't cleared on each render, just updated.

https://github.com/kalkih/mini-graph-card/blob/709a5ded6fec03501424a15b534f5c733742f00f/src/main.js#L811-L843

https://github.com/kalkih/mini-graph-card/blob/709a5ded6fec03501424a15b534f5c733742f00f/src/main.js#L564-L580

PR #305, opened 2020-03-17
update Readme with correct default value

optama:
According to buildConfig.js default value for height in Card options is 100 (not 150 as currently stated in Readme).


kalkih:
Thanks!

kalkih:
Could you change the text to `Set a custom height of the graph.` while you're at it? It's applieed to both line and bar graphs.
Thanks!

PR #306, opened 2020-03-17
changed description of height property to also include bar graphs

optama:


kalkih:
Please do this change in #305, thanks.

PR #318, opened 2020-03-31
Adaptation to allow for rendering a logarithmic graph. 

alonjr:
Hey karl, I had a need so did a small change. i've seen that others were seeking something similar but without any luck finding a solution.

For me this was useful for data sets that fluctuate over orders of magnitude, e.g. luminance readings from a light sensor (1 to 64k lux).

I added "logarithmic" attribute in ui-lovelace.yaml.

I only tested for line graphs, but if this is of interest, i'm happy to spend the time to make this a legit feature.

what do you think?

### DIFF
--- a/src/graph.js
+++ b/src/graph.js
@@ -5,7 +5,7 @@ import {
 } from './const';
 
 export default class Graph {
-  constructor(width, height, margin, hours = 24, points = 1, aggregateFuncName = 'avg', groupBy = 'interval', smoothing = true) {
+  constructor(width, height, margin, hours = 24, points = 1, aggregateFuncName = 'avg', groupBy = 'interval', smoothing = true, logarithmic = false) {
     const aggregateFuncMap = {
       avg: this._average,
       max: this._maximum,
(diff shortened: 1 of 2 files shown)


kalkih:
Hello, thanks for the PR.

This would be cool, there are some issues with the implementation, log10(0) equals -infinity and log10(negative value) equals NaN, so this would only work if the min, max and all graph values are positive otherwise the graph will break.

alonjr:
Karl, Thanks for your feedback. I’ll definitely go through the edge cases and get back to you.

Cheers
Alon

Alon Rosenfeld | alon@vayapost.com | +34 634588818

awirtz:
You could artificially cap the values and avoid getting into negative/ NaN territory like this:

const max = this._logarithmic ? Math.log10(Math.max(1, this.max)) : this.max;
const min = this._logarithmic ? Math.log10(Math.max(1, this.min)) : this.min;

and

const val = this._logarithmic ? Math.log10(Math.max(1, coord[V])) : coord[V];

That would somewhat imply that people know what they are doing when using the logarithmic scale, otherwise they might complain that the graph is doing funny stuff with their numbers...

Can you please continue on this? The graphs of my lux-sensors are not helpful at the moment...

yuvalabou:
Would love to see that being merge for my power monitoring graph!

Hypfer:
Something like this would be very useful for light sensor measurements

da-anda:
@kalkih I suppose this PR can be closed

Issue #326, opened 2020-04-09
Timestamp issue?


Flight777:
Hi there, first of all, great work on this card, it makes HASS so much nicer! :)

Forgive me if I'm asking something stupid, but need some help in troubleshooting this (or it might be a small bug in the end):

I have a bar chart that shows my actual energy cost, based on DSMR Reader integration (via MQTT).

```
aggregate_func: last
animate: true
cache: false
entities:
  - entity: sensor.dsmr_day_consumption_total_cost
group_by: date
hours_to_show: 168
name: Energy Cost
show:
  graph: bar
type: 'custom:mini-graph-card'
```
The total cost is accumulating (or reducing based on my solar panel production) constantly, thus I only want it to display the last value (and historically the last value of the day) to be accurate.

Now here is where mini-graph behaves a bit erratic, 2/3 times it shows 0 for all the historic days, which is the first value that the DSMR Reader Posts for the new day at 0:00. 1/3 times it does show the correct, last value of that day, posted at 23:59:xx. It seems somehow mini-graph might have some issues with selecting the correct timestamp if they are very close together, and then takes the first value of the new day, as the "last" value of the old day (hope that makes sense).

kalkih:
Hello!
Does the correct value show up in the displayed state, just not correctly in the graph?
This card by default applies smoothing to graph values, which kind of is a problem when using "last & first" aggregation.
Try setting `smoothing: false`

Is it possible that the history entries are delayed and not updated until after 23:59:59 ?
You should be able to check this with the default history-graph by hovering over values.

Flight777:
Hi,

Tried the smoothing just now but did not fix the behaviour.

Yes the state is displayed correctly in the historic graphs.

Timestaps are +/- 1 min apart, so I checked for a few days and it's always similar to:

![image](https://user-images.githubusercontent.com/5236143/78884952-0a809400-7a5c-11ea-94f0-f3e8c99f9d6e.png)

![image](https://user-images.githubusercontent.com/5236143/78885004-2126eb00-7a5c-11ea-9332-d082ec1a7d4f.png)

![image](https://user-images.githubusercontent.com/5236143/78885072-3b60c900-7a5c-11ea-9818-a049896807ac.png)

kalkih:
Okay, thanks for the details.

jaguba:
[...]

Issue #289, opened 2020-02-29
Wrong last bar value - always zero.


dzikus:
Last bar on graph is wrong, also hours to show omitting last day (history-graph show one day more with 720 hours).

![image](https://user-images.githubusercontent.com/3209066/75612285-54539180-5b22-11ea-885b-a9dbc7c0b1a3.png)
![image](https://user-images.githubusercontent.com/3209066/75612286-57e71880-5b22-11ea-8ff8-0269a343a715.png)
￼
```yaml
aggregate_func: max
cache: false
color_thresholds:
  - color: '#f0f042'
    value: 15
  - color: '#f0e342'
    value: 7.5
  - color: '#d0c442'
    value: 5
  - color: '#c0c042'
    value: 0
entities:
  - entity: sensor.daily_pv4
    index: 0
    name: Dzienna produkcja PV
group_by: date
hours_to_show: 720
icon: 'mdi:solar-power'
show:
  graph: bar
type: 'custom:mini-graph-card'
```

```yaml
entities:
  - entity: sensor.daily_pv4
hours_to_show: 720
refresh_interval: 0
type: history-graph
```

kalkih:
Group by date will often show one less day than the history graph since using "group_by: date" will only display whole days compared to history graph where the first entry will be current time of day 720h. With mini-graph-card you'll get end of current day (00:00) - 720h.

I need more information than "last bar in graph is wrong", I can't really do anything with that information.

dzikus:
It is showing 0, but there is 2.3 value as I show on history-graph card. Bellow again there is 248 max value on history-graph but on mini-graph-card there is zero 0.


<img width="344" alt="Zrzut ekranu 2020-03-1 o 11 43 54" src="https://user-images.githubusercontent.com/3209066/75624136-f078a980-5bb1-11ea-93dd-e601bf648b5a.png">
<img width="342" alt="Zrzut ekranu 2020-03-1 o 11 47 38" src="https://user-images.githubusercontent.com/3209066/75624196-78f74a00-5bb2-11ea-93d7-3fbc439c8750.png">

```yaml
aggregate_func: max
cache: false
entities:
  - color: '#0399ff'
    entity: sensor.daily_water
    index: 0
    name: Example of last bar with 0 value
group_by: date
hours_to_show: 168
icon: 'mdi:water-pump'
show:
  graph: bar
type: 'custom:mini-graph-card'
```

On my every bar graph last bar is 0:
[...]

PR #327, opened 2020-04-09
Fixed incorrect value for the first point

maxwroc:
Hello again @kalkih,

I have noticed one issue. You were not resetting seconds and milliseconds when the endTime was set. This was resulting in slightly wrong value buckets. 

This wouldn't be a big deal alone but you have this condition:
`if (coords[0] && coords[0].length) coords[0] = [coords[0][coords[0].length - 1]];`

For me it looks little bit hacky way of solving some specific scenario. :) IMO we should think about changing this as it works for me only because I care about max value which is always at the end of the bucket. If someone would like to take min or avg this won't work.

Example case which was failing for me:
image

And the result:
image

Take care buddy. I'm impressed how committed you are for these mini-* cards (I see update after update hehe). I use both of them! Thanks for your hard work!

Max

### DIFF
--- a/src/graph.js
+++ b/src/graph.js
@@ -218,7 +218,7 @@ export default class Graph {
     switch (this._groupBy) {
       case 'date':
         this._endTime.setDate(this._endTime.getDate() + 1);
-        this._endTime.setHours(0, 0);
+        this._endTime.setHours(0, 0, 0, 0);
         break;
       case 'hour':
         this._endTime.setHours(this._endTime.getHours() + 1);


maxwroc:
It fixes #289

kalkih:
Great spotted @maxwroc.
Thanks for the PR, possibly fixes https://github.com/kalkih/mini-graph-card/issues/326 as well?

> This wouldn't be a big deal alone but you have this condition:
> `if (coords[0] && coords[0].length) coords[0] = [coords[0][coords[0].length - 1]];`

Indeed that solution is a little hacky and shouid be addressed 😆

Thanks Max!

Issue #288, opened 2020-02-28
Show only differences


Dario82:
Hello, from my air conditioner I have this situation:

from 08.00 to 9.00 -> 8.5kwh
from 9.00 to 10.00 -> 9.9kwh (8.5 + 1.4)
from 10.00 to 11.00 -> 10.7kwh (9.9 + 0.8)
from 11.00 to 12.00 -> 11.45kwh (10.7 + 0.75)

I've created a bar graph and I have this:
![image](https://user-images.githubusercontent.com/39548149/75542889-f68b5080-5a20-11ea-8f9a-0080be060340.png)

But I want the differences between hours (0, 1.4, 0.8, 0.75)
Is it possible?

Thank you.

Dario82:
This is my configuration:

  - type: 'custom:mini-graph-card'
    aggregate_func: max
    entities:
      - entity: sensor.daily_power
        index: 0
        name: kWh
      - entity: sensor.daily_costs
        index: 1
        name: €
    group_by: hour
    points_per_hour: 0.25
    hour24: true
    cache: false
    hours_to_show: 24
    icon: 'mdi:contrast-box'
    name: Consumo orario
    show:
      average: true
      extrema: true
      fill: false
      graph: bar
      points: true

JohNan:
:+1:
I would also have use of this kind of graph.

kalkih:
This should already be possible by setting up a utility sensor, see [utility sensor](https://www.home-assistant.io/integrations/utility_meter/) and https://github.com/kalkih/mini-graph-card/issues/197

Issue #197, opened 2019-11-04
Feature request: aggregate_func: diff


orderofcodeadmin:
Hi,
This feature request is not that big of a deal but it would be very useful for counters presentation.
Graph should display difference between last values of groups.
Useful can be aggregation like "first" and "last" too.

```  
  - type: custom:mini-graph-card
    aggregate_func: diff
    group_by: date
    entities:
       - entity: sensor.gas_counter
```

kalkih:
Hey,
Please elaborate on what you mean with diff, how should it be calculated, preferably with an example.
Thanks!

orderofcodeadmin:
Many sensors related to gas, energy, water will return just counter (accumulated usage).
100,101|104,105,106,107,108,109|110,111,112|113|... the values are stores at a certain time
grouping by time (hour/day/month) we calculate differences, ('|' represents time intervals)
So I expect values 1, 8 (109-101), 3 (112 - 109), 1 (113-112)
First value of series is little problem because you don't know what you should use 0 or first value.
If data exist you can ask about previous values, but if it is begin of data is should be first value or additional parameter for user.
In that simple way graph can present per time range,  for example usage per hour/day, without "utility_meter"

I hope I described it quite clearly.

kalkih:
Thanks for the explanation, I see what you mean now.
I don't really feel like this is worth spending time on when we already have a perfectly working solution in the utility meter component.

elupus:
I would just like to add my +1 for this. I would not consider the utility meter perfectly working. It just adds a sensor entity for the same data, where you pre.hand need to know how you want to present the data.

when you realize you want per-month differences, you need to wait several months get any type of data graphed. The data is already in the database, but you can not use it. The utility_meter is in my opinion severely flawed from a presentation perspective.

Also with a diff aggregation (or might want to call it a timeshift difference), you can have smooth monthly running average of your consumption, since it doesn't need fixed breakpoint.

danielnguyen:
@kalkih Any chance we can update master or re-tag (even as a beta tag) to be able to take advantage of PR #329?

PR #329, opened 2020-04-10
Add delta option in aggregate_func

PierreB49:
This option allows to show delta for instance to show daily or hourly consumption from cumulative data.

### DIFF
--- a/src/graph.js
+++ b/src/graph.js
@@ -25,6 +25,7 @@ export default class Graph {
     this._min = 0;
     this.points = points;
     this.hours = hours;
+    this.aggregateFuncName = aggregateFuncName;
     this._calcPoint = aggregateFuncMap[aggregateFuncName] || this._average;
     this._smoothing = smoothing;
     this._groupBy = groupBy;
@@ -215,7 +216,7 @@ export default class Graph {
   }
 
   _lastValue(items) {
-    if (this._calcPoint(items) === this._delta(items)) {
+    if (this.aggregateFuncName === 'delta') {
       return 0;
     } else {
       return parseFloat(items[items.length - 1].state) || 0;
--- a/src/graph.js
+++ b/src/graph.js
@@ -49,18 +49,20 @@ export default class Graph {
     if (!this._history) return;
     this._updateEndTime();
 
-    const coords = this._history.reduce((res, item) => this._reducer(res, item), []);
+    const histGroups = this._history.reduce((res, item) => this._reducer(res, item), []);
 
     // drop potential out of bound entry's except one
-    if (coords[0] && coords[0].length) coords[0] = [coords[0][coords[0].length - 1]];
+    if (histGroups[0] && histGroups[0].length) {
+      histGroups[0] = [histGroups[0][histGroups[0].length - 1]];
+    }
 
     // extend length to fill missing history
     const requiredNumOfPoints = Math.ceil(this.hours * this.points);
-    coords.length = requiredNumOfPoints;
+    histGroups.length = requiredNumOfPoints;
 
-    this.coords = this._calcPoints(coords);
-    this.min = Math.min(...this._calcPoints(coords).map(item => Number(item[V])));
-    this.max = Math.max(...this._calcPoints(coords).map(item => Number(item[V])));
+    this.histGroups = this._calcPoints(histGroups);
+    this.min = Math.min(...this.histGroups.map(item => Number(item[V])));
+    this.max = Math.max(...this.histGroups.map(item => Number(item[V])));
   }
 
   _reducer(res, item) {
@@ -226,6 +228,10 @@ export default class Graph {
   _updateEndTime() {
     this._endTime = new Date();
     switch (this._groupBy) {
+      case 'month':
+        this._endTime.setMonth(this._endTime.getMonth() + 1);
+        this._endTime.setDate(1);
+        break;
       case 'date':
         this._endTime.setDate(this._endTime.getDate() + 1);
         this._endTime.setHours(0, 0);
--- a/src/graph.js
+++ b/src/graph.js
@@ -13,6 +13,7 @@ export default class Graph {
       first: this._first,
       last: this._last,
       sum: this._sum,
+      delta: this._delta,
     };
 
     this._history = undefined;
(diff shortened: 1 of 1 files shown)


src/graph.js
```diff
@@ -57,8 +58,8 @@ export default class Graph {
     coords.length = requiredNumOfPoints;
 
     this.coords = this._calcPoints(coords);
-    this.min = Math.min(...this.coords.map(item => Number(item[V])));
-    this.max = Math.max(...this.coords.map(item => Number(item[V])));
+    this.min = Math.min(...this._calcPoints(coords).map(item => Number(item[V])));
```
maxwroc, line 61: Can you explain why do you need to calculate points again?

src/graph.js
```diff
@@ -57,8 +58,8 @@ export default class Graph {
     coords.length = requiredNumOfPoints;
```
maxwroc, line 58: Can you change the name of this var as currently it is confusing e.g.: coords => groupedHist (or histGroups)

src/graph.js
```diff
@@ -209,8 +210,16 @@ export default class Graph {
     return items.reduce((sum, entry) => sum + parseFloat(entry.state), 0);
   }
 
+  _delta(items) {
+    return this._maximum(items) - this._minimum(items);
+  }
+
   _lastValue(items) {
-    return parseFloat(items[items.length - 1].state) || 0;
+    if (this._calcPoint(items) === this._delta(items)) {
```
maxwroc, line 218: Can you please explain this condition? Why for delta the last value should always be 0?

I think it is better to store aggregate func name as property and do the check here which is more clear IMO. Now we have redundant computation which is not clear (adding a comment would be useful if eventually you'll leave it)

PierreB49:
> Hey, before addressing my comments please wait for @kalkih opinion.
> 
> Nice and simple extension please update documentation and give the example use case there (maybe update/mention this in the [energy usage example](https://github.com/kalkih/mini-graph-card#grouping-by-date))
> 
> edit:
> BTW I don't know how come I didn't came up with this solution (my first change for this lib was actually to give a way to show the energy daily consumption). I think I was too focused on the getting the results form utility meter (not direct results from the sensor) :) Good job!

Thanks you for your reply. I'm a beginner in nodejs so there might be room for improvement in my code. I adapted the card for my usage and thought it could be useful for someone else.

src/graph.js
PierreB49: These min and max values are used to compute the scale of the bar graph. If you use the initial data but not the values transformed thru the "aggregate_func" there is a scale issue. It's not too big of a deal with the min/max/avg/first/last but  with the delta function the scale is way to small.

src/graph.js
PierreB49: Honestly i'm not very proud of that code. Your solution seems way better.

If the sensor value are not updated anymore (for instance the meter is down) there is no new value. In this cas with the initial code the selected value is the last value of the meter. With the delta function if there is no updated value it should show 0 (no change in value).

src/graph.js
maxwroc: Maybe I'm missing something... byt sorry, I still don't understand why it is required. In line 60 you call `this._calcPoints(coords)` so why do you need to repeat it in both lines below if the input collection doesn't change. 

Call to this function with the same input collection should produce exactly the same result every time.

maxwroc:
> > Hey, before addressing my comments please wait for @kalkih opinion.
> > Nice and simple extension please update documentation and give the example use case there (maybe update/mention this in the [energy usage example](https://github.com/kalkih/mini-graph-card#grouping-by-date))
> > edit:
> > BTW I don't know how come I didn't came up with this solution (my first change for this lib was actually to give a way to show the energy daily consumption). I think I was too focused on the getting the results form utility meter (not direct results from the sensor) :) Good job!
> 
> Thanks you for your reply. I'm a beginner in nodejs so there might be room for improvement in my code. I adapted the card for my usage and thought it could be useful for someone else.

And it is awesome that you have decided to share the solution and update the lib! 

One thing worth to mention additionally in the documentation is that it may not give exactly correct  results. What I mean here is that in theory we should store somewhere the last state from previous time bucket and calc delta to the value from the next one. Here you calculate delta within single time bucket which in some corner case situations may produce incorrect results. 
But anyway I think for majority of people this won't be a problem so I'm fine with this approach as the implementation is simple and does the job. It will be useful for people who don't want to add another sensor - [utility meter](https://www.home-assistant.io/integrations/utility_meter/).

I wonder what is @kalkih's optionion. I remember he didn't want to add similar functionality as it can be achieved via utility meter.

kalkih:
Hello, this is indeed a nice addition to the card, thanks!
As Max said I've pointed people to utility meter when they've asked for functionality like this but I'm ok with this solution it's simple & could work fairly well in some configurations.

Just address the comments made by Max and add it to the documentation and I'll merge this, thank you!

src/graph.js
PierreB49: You're right ! I updated the code in commit e7f5100

src/graph.js
PierreB49: Updated in commit e7f5100

src/graph.js
```diff
@@ -47,18 +49,20 @@ export default class Graph {
...
 
     // extend length to fill missing history
     const requiredNumOfPoints = Math.ceil(this.hours * this.points);
-    coords.length = requiredNumOfPoints;
+    histGroups.length = requiredNumOfPoints;
 
-    this.coords = this._calcPoints(coords);
-    this.min = Math.min(...this.coords.map(item => Number(item[V])));
-    this.max = Math.max(...this.coords.map(item => Number(item[V])));
+    this.histGroups = this._calcPoints(histGroups);
```
maxwroc, line 63: You have edited little bit too much now ;) The local var name which I have asked to change was confusing because it wasn't containing graph coords and the name was exactly the same as the class property (so the impression was that the type of the value was the same - which was not true). 
Class property `this.histGroups` doesn't contain grouped history values (as name suggests) but actually calculated graph points/coords. Please leave here the old name.

maxwroc:
This probably covers requests: https://github.com/kalkih/mini-graph-card/issues/197, https://github.com/kalkih/mini-graph-card/issues/288

maxwroc:
@kalkih can you merge this please? my last comment is a minor thing which we can do in another PR

kalkih:
Great job, thanks for the contribution!

Needs to be added to the docs before next release.

maxwroc:
Ok. Let me do it

maxwroc:
https://github.com/kalkih/mini-graph-card/pull/332

Issue #114, opened 2019-07-04
Feature Request: a "soft" upper bound


worgarside:
I've got a couple of graphs where I'd like to set an upper bound, but the value being plotted does sometimes go over that value and then it doesn't look as nice. An example being a step count: the upper bound is 10,000 (so I can see how close I am to that goal), but sometimes I'll do more than 10K steps a day and then it'll go off the graph.

I was thinking that a nice solution to this would be the ability to make the upper bound "soft". i.e. It stays at the set value Y, until the value on the graph (Y1) is greater than Y, at which point the upper bound becomes Y1 so all values are plotted.

kalkih:
Hmm, that's actually a neat idea, I like it!
Should be fairly easy to implement as well.

michaelblight:
One option that doesn't involve a new feature would be to also store your step goal in Home Assistant and plot that too. With a step goal of 10,000 it will plot a horizontal line at this point, which will also be the maximum if the steps achieved is lower. However, if the steps achieved exceeds 10,000, they will be the new maximum for the graph. i.e. no need to set an upper bound.
You also get the benefit that you can change your step goal over time easily, and could even have different values for the weekends (for example).

worgarside:
Ah that's actually a good shout! Cheers for the idea, I'll definitely implement that until I have time to look at adding a new feature

worgarside:
For reference, this is my LoveLace config:

```yaml
- type: custom:mini-graph-card
  points_per_hour: 6
  icon: mdi:shoe-print
  entities:
    - entity: sensor.daily_step_count
    - entity: sensor.daily_step_count_goal
      color: '#00000000'
  lower_bound: 0
  line_width: 3
  show:
    points: false
    legend: false
  color_thresholds:
    - value: 0
      color: '#ff0000'
    - value: 10000
      color: '#00ff00'
```

and this is my goal template sensor:
```yaml
daily_step_count_goal:
  friendly_name: 'Daily Step Count Goal'
  unit_of_measurement: 'Steps'
  value_template: "{{ 10000 }}"
```

By setting the color of the goal sensor to transparent (`#00000000`) and hiding the legend on the graph, the graph doesn't actually look any different! 🎉

PR #330, opened 2020-04-14
soft_bound config

Silencer2K:
A new config parameter that allows to extend the boundaries when graph is out of range

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,10 +93,10 @@ This card is available in [HACS](https://github.com/custom-components/hacs/issue
 | align_header | string | `default` | v0.2.0 | Set the alignment of the header, `left`, `right`, `center` or `default`.
 | align_icon | string | `right` | v0.2.0 | Set the alignment of the icon, `left`, `right` or `state`.
 | align_state | string | `left` | v0.2.0 | Set the alignment of the current state, `left`, `right` or `center`.
-| lower_bound | number |  | v0.2.3 | Set a fixed lower bound for the graph Y-axis.
-| upper_bound | number |  | v0.2.3 | Set a fixed upper bound for the graph Y-axis.
-| lower_bound_secondary | number |  | v0.5.0 | Set a fixed lower bound for the graph secondary Y-axis.
-| upper_bound_secondary | number |  | v0.5.0 | Set a fixed upper bound for the graph secondary Y-axis.
+| lower_bound | number *or* string |  | v0.2.3 | Set a fixed lower bound for the graph Y-axis. String value starting with ~ (e.g. `~50`) specifies soft bound.
+| upper_bound | number *or* string |  | v0.2.3 | Set a fixed upper bound for the graph Y-axis. String value starting with ~ (e.g. `~50`) specifies soft bound.
+| lower_bound_secondary | number *or* string |  | v0.5.0 | Set a fixed lower bound for the graph secondary Y-axis. String value starting with ~ (e.g. `~50`) specifies soft bound.
+| upper_bound_secondary | number *or* string |  | v0.5.0 | Set a fixed upper bound for the graph secondary Y-axis. String value starting with ~ (e.g. `~50`) specifies soft bound.
 | smoothing | boolean | `true` | v0.8.0 | Whether to make graph line smooth.
 | state_map | [state map object](#state-map-object) |  | v0.8.0 | List of entity states to convert (order matters as position becomes a value on the graph).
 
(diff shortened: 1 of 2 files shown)


src/main.js
```diff
@@ -730,10 +730,14 @@ class MiniGraphCard extends LitElement {
...
-        ? config.lower_bound
+        ? !config.soft_bound
+          ? config.lower_bound
+          : Math.min(config.lower_bound, ...this.primaryYaxisSeries.map(ele => ele.min))
         : Math.min(...this.primaryYaxisSeries.map(ele => ele.min)) || this.bound[0],
       config.upper_bound !== undefined
-        ? config.upper_bound
+        ? !config.soft_bound
+          ? config.upper_bound
+          : Math.max(config.upper_bound, ...this.primaryYaxisSeries.map(ele => ele.max))
```
maxwroc, line 740: Few things:
* double inline conditions are evil ;), please refactor
* you have added this option for main axis please do the same for secondary (to have consistency), I think you don't need a separate setting for it and you can reuse this one
* please update documentation

If I understand correctly you want to have a fixed bound but in case of value out of range we should render it correctly, righ?

@kalkih this seems to me like a bugfix - not a new feature, no? I'm trying to think about scenario where such hard bounds are useful but TBH nothing reasonable comes to my mind

kalkih: I think it could potentially be confusing for the user if the bound is specifically set but then also is dynamic, so I like the idea of having this as an option.
While I too can't find a use case where this isn't preferred, I'm sure there are, so we should keep that option available.

If it's documented well we could probably make this enabled by default though.

kalkih:
Closes: #114

src/main.js
```diff
@@ -729,21 +729,29 @@ class MiniGraphCard extends LitElement {
...
+      config.lower_bound_secondary === undefined
+        ? Math.min(...this.secondaryYaxisSeries.map(ele => ele.min)) || this.boundSecondary[0]
+        : config.soft_bounds_secondary
+          ? Math.min(config.lower_bound_secondary, ...this.secondaryYaxisSeries.map(ele => ele.min))
+          : config.lower_bound_secondary,
+      config.upper_bound_secondary === undefined
+        ? Math.max(...this.secondaryYaxisSeries.map(ele => ele.max)) || this.boundSecondary[1]
+        : config.soft_bounds_secondary
+          ? Math.max(config.upper_bound_secondary, ...this.secondaryYaxisSeries.map(ele => ele.max))
+          : config.upper_bound_secondary,
```
maxwroc, line 754: Sorry but I sustain my opinion: nested inline conditions make code not clear and hard to read. Please change it to regular `if` statements.

Another idea came up to my mind now: what about not adding new setting entry but allow values for bounds like: `~0`. The `~` prefix would mean that this is not hard boundary . What do you think?

I would do it in the following way:
```js
this.bound = [
    this.getBoudary("min", config.lower_bound, this.primaryYaxisSeries, this.bound[0]),
    this.getBoudary("max", config.upper_bound, this.primaryYaxisSeries, this.bound[1]),
];
```

```js
getBoudary(type, configVal, series, defaultVal) {
    if (configVal === undefined) {
        // dynamic boundary depending on values
        return Math[type](...series.map(ele => ele[type])) || defaultVal
    }

    if (configVal[0] != "~") {
        // fixed boundary
        return configVal;
    }

    // soft boundary (respecting out of range values)
    configVal = parseInt(configVal.substr(1));
    Math[type](configVal, ...series.map(ele => ele[type]));
}
```

README.md
```diff
@@ -95,8 +95,10 @@ This card is available in [HACS](https://github.com/custom-components/hacs/issue
 | align_state | string | `left` | v0.2.0 | Set the alignment of the current state, `left`, `right` or `center`.
 | lower_bound | number |  | v0.2.3 | Set a fixed lower bound for the graph Y-axis.
 | upper_bound | number |  | v0.2.3 | Set a fixed upper bound for the graph Y-axis.
+| soft_bounds | boolean | `false` |  | Allow to extend fixed bounds when graph is out of range.
```
maxwroc, line 98: As @kalkih suggested it might be true by default

src/main.js
Silencer2K: Good idea. Now config looks much better: [Screenshot](https://s.mail.ru/KW4e/tG6Cpzzdy)

kalkih:
Looks good,  good idea with the `~` prefix, great job you two.

ZeevoX:
It should probably be made clear in the README from which version onwards soft bounds are supported. As a new user I had no idea that soft bounds were newly introduced because it says from v0.2.3 and was wondering why they weren't working.

PR #356, opened 2020-05-20
Added clarification as to what "HACS" is

EpicLPer:
I did not know what HACS is and at first thought it's the Official Home Assistant Community store. Since the HACS link only pointed to the "Issues" page and didn't describe what HACS is at all I thought it'd only be fair to describe it a little further for people (like me) that just stumbled across this for their first time :)


kalkih:
Great, thanks!

PR #357, opened 2020-05-21
[WIP] Adding aspect ratio support

mcpowell87:
Opening up a WIP to open discussion about adding support for an aspect_ratio config, which will allow the user to force the card to render in a specified "aspect ratio".  From personal experience and from browsing the community forums, I know this is a desired feature by some because it allows users to achieve a conformed look when using the mini-graph card in a horizontal stack with other cards (especially those that already support aspect ratios, such as custom button-card).

This is a basic implementation that sizes the card according to the aspect_ratio config (in the format of 1/1, 1/2, 2/1, etc).  It works similarly to how the aspect_ratio works in the custom button-card.  Excluding the aspect_ratio config will give back the cards ability to dynamically size itself.  The caveat is that it does not, in it's current state, account for the overflow of elements within the card (e.g, enabling extrema, icon, and state when your card is sufficiently small because it's in a horizontal stack with 3 other cards will cause some information to be hidden).

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -1,5 +1,4 @@
 import { LitElement, html, svg } from 'lit-element';
-import { styleMap } from 'lit-html/directives/style-map';
 import localForage from 'localforage/src/localforage';
 import Graph from './graph';
 import style from './style';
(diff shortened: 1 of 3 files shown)


src/main.js
```diff
@@ -169,21 +170,32 @@ class MiniGraphCard extends LitElement {
...
-        ?gradient=${config.color_thresholds.length > 0}
-        ?hover=${config.tap_action.action !== 'none'}
-        style="font-size: ${config.font_size}px;"
-        @click=${e => this.handlePopup(e, config.tap_action.entity || this.entity[0])}
-      >
-        ${this.renderHeader()} ${this.renderStates()} ${this.renderGraph()} ${this.renderInfo()}
-      </ha-card>
+      <div 
+        id="aspect-ratio"
+        style="${styleMap(aspectRatio)}">
```
maxwroc, line 183: TBH I'm not convinced that this is worth to add a new dependency

mcpowell87: You're absolutely right about that, good catch.  I removed that dependency.

PR #362, opened 2020-05-25
Added value_factor support, to scale value shown in state label

dzikimarian:
Hey,
My router reports everything in bytes, which is not very handy for daily use, therefore I added ability to scale value to other unit. If you like it, fell free to merge it.

obraz
Default settings

obraz

Value factor applied
```
value_factor: -9
decimals: 2
unit: GB
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,7 @@ HACS is a 3rd party Community Store, NOT the one you can already find under "Sup
 | upper_bound_secondary | number *or* string |  | v0.5.0 | Set a fixed upper bound for the graph secondary Y-axis. String value starting with ~ (e.g. `~50`) specifies soft bound.
 | smoothing | boolean | `true` | v0.8.0 | Whether to make graph line smooth.
 | state_map | [state map object](#state-map-object) |  | v0.8.0 | List of entity states to convert (order matters as position becomes a value on the graph).
+| value_factor | number | 0 | vX.X.X | Scale value by order of magnitude (e.g. convert Watts to kilo Watts), use negative value to scale down. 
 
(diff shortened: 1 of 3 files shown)


kalkih:
Thanks, great addition!

clavdal:
Hi. Great feature. In my case, I need to convert windspeed from KM/H to M/S - so have to multiply it by 0,277778 - This feature doesn't support that if I understand it correctly  - Is any other way to make such a modification

ildar170975:
> Is any other way to make such a modification

template sensor

dzikimarian:
Well, there's a way :-) Since value_factor effectively raises 10 to the power of its value, then it's matter of question: "to which power I need raise 10 in order to get 0.27778“.

This can be done by calculating log10(0.27778), and the answer is -0,556302153 - set value_factor to it and it should be fine (I calculated value over morning coffee, so please double check it's valid :-))

clavdal:
Thanks - Both solutions seem to work - I will save the Template sensor for the next challenge.

ildar170975:
In some cases converting units a not a task for a custom card.
If some integration provides a value with inappropriate unit - shall we convert it in the card?
For instance, Xiaomi Miio gives a total duration of cleaning for robot vacuums in seconds - which is inconvenient.
Surely a user have to create a template sensor to convert this value to "hours" (or whatever) - and then this converted value may be displayed both in Entities card & any graph card (w/o a need to convert).
Alternatively, if some value is displayed with an original unit in many cards and you want to change a uniit for just ONE card - then yes, you are interested to convert in in the card.

PR #397, opened 2020-07-11
Example images before code

agneevX:
This PR changes the way examples are shown in the README.

Images before the code should make readability easier.


kalkih:
Makes sense, thanks!

PR #408, opened 2020-07-27
Allow setting of min range for Y-axis

tal:
If there's small variance in the dataset small changes can give large swings in the graph. When you set a minimum range you can make sure that if there's small variance you get a fairly flat curve. This PR adds a setting for that on both Y axis.

Tested in browser and passes linter.

With min range set to 10:
image

Without it set:
image

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -730,6 +730,10 @@ class MiniGraphCard extends LitElement {
   }
 
   getBoundary(type, configVal, series, defaultVal) {
+    if (!(type in Math)) {
+      throw new Error(`The type "${type}" is not present on the Math object`);
+    }
+
     if (configVal === undefined) {
       // dynamic boundary depending on values
       return Math[type](...series.map(ele => ele[type])) || defaultVal;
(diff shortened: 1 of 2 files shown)


src/main.js
```diff
@@ -744,13 +748,42 @@ class MiniGraphCard extends LitElement {
...
+      const currentRange = Math.abs(this.bound[0] - this.bound[1]);
+      const diff = minBoundRange - currentRange;
+
+      // Doesn't matter if minBoundRange is NaN because this will be false if so
+      if (diff > 0) {
+        this.bound = [
+          this.bound[0] - diff / 2,
+          this.bound[1] + diff / 2,
+        ];
+      }
```
kalkih, line 766: Should make this into a function, since it's basically the same code repeated over again for the secondary axis.
Would also be useful if we decide to add support for more/unlimited axes in the future.

tal: I felt like the increase in complexity to factor this out was much higher than the increased utility. If you support unlimited values it'd make sense to do this then. There would likely be a lot of changes associated with that change.

kalkih:
Thanks for the PR, great feature 👍

Issue #424, opened 2020-08-21
Minifying the resulting JS file more


Fusseldieb:
Hi,
I just saw that in the resulting JS file there are a lot of white spaces, many of them probably unecessary, wasting bandwidth, and every KB saved is a great thing! Doesn't sound like much, but still...

I searched for "at least" double spaces in VSCode and then deleted them to compare the sizes.

Before:
<img src="https://user-images.githubusercontent.com/4715129/90891593-57716100-e392-11ea-9d7d-2aa602893c50.png" width=400>

After:
<img src="https://user-images.githubusercontent.com/4715129/90891611-5dffd880-e392-11ea-94cc-bfca164e7828.png" width=400>

Please note that this might not be 100% accurate, nor have I tested the code for functionality, but most of these spaces could be eliminated without any issue, especially those between html tags, before/after line breaks `\n`) and such.

I'm not just pointing it out. I want to help too. That means if I'm allowed I would clone this repo and include a little [grunt script](https://gruntjs.com/) into it. This script would essentially take the final JS file, erase the spaces following a regex or something and then overwrite it. There are probably other ways, but I was thinking that this would be the straightest route.

EDIT: I just saw that ESLint has some minifying options too, but I'm not familiar with it, I just flown over the related documentation real quick.

Fusseldieb:
> Perhaps https://github.com/cfware/babel-plugin-template-html-minifier ?

Thanks for the suggestion! I tried it, included it into `.babelrc` following [the example for lit-element](https://github.com/cfware/babel-plugin-template-html-minifier#usage), but it didn't minify a thing. Spend coutless hours trying to adjust parameters, but still nothing.

kalkih:
Terser has a [rollup plugin](https://www.npmjs.com/package/rollup-plugin-terser) so if that does the trick I'm all for it 👍

Fusseldieb:
> Terser has a [rollup plugin](https://www.npmjs.com/package/rollup-plugin-terser) so if that does the trick I'm all for it 👍

I used it according to documentation, but for some reason it just removed the comments from the final file, leaving the JS file with all whitespaces in place, making it even bigger. I thought a plugin was interfering with it, so I disabled everything and tried to `rollup-plugin-terser` almost barebone, to no avail.

[...]

PR #435, opened 2020-08-31
Minified even more with Terser

Fusseldieb:
As discussed in #424, gives a dist file reduction up to 7KB. 

Build tested and working.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,8 @@
   "dependencies": {
     "@kalkih/lz-string": "^1.4.5",
     "lit-element": "^2.2.1",
-    "localforage": "^1.7.3"
+    "localforage": "^1.7.3",
+    "rollup-plugin-minify-html-literals": "^1.2.5"
   },
   "devDependencies": {
     "@babel/cli": "^7.6.4",
@@ -42,10 +43,12 @@
     "rollup-plugin-node-resolve": "^3.4.0"
   },
   "scripts": {
-    "build": "npm run lint && npm run rollup && npm run babel",
+    "build": "npm run lint && npm run rollup && npm run babel && npm run terser",
+    "terser": "terser dist/mini-graph-card-bundle.js -o dist/mini-graph-card-bundle.js --mangle --comments false --config-file terser.config.json",
     "rollup": "rollup -c",
     "babel": "babel dist/mini-graph-card-bundle.js --out-file dist/mini-graph-card-bundle.js",
     "lint": "eslint src/* --ext .js",
+    "lint:fix": "npm run lint -- --fix",
     "watch": "rollup -c --watch"
   }
 }
--- a/package.json
+++ b/package.json
@@ -40,7 +40,8 @@
     "eslint-config-airbnb-base": "^13.1.0",
     "eslint-plugin-import": "^2.17.2",
     "rollup": "^0.66.6",
-    "rollup-plugin-node-resolve": "^3.4.0"
+    "rollup-plugin-node-resolve": "^3.4.0",
+    "terser": "^5.2.1"
   },
   "scripts": {
     "build": "npm run lint && npm run rollup && npm run babel && npm run terser",
--- a/package.json
+++ b/package.json
@@ -21,8 +21,7 @@
   "dependencies": {
     "@kalkih/lz-string": "^1.4.5",
     "lit-element": "^2.2.1",
-    "localforage": "^1.7.3",
-    "rollup-plugin-minify-html-literals": "^1.2.5"
+    "localforage": "^1.7.3"
   },
   "devDependencies": {
     "@babel/cli": "^7.6.4",
@@ -40,6 +39,7 @@
     "eslint-config-airbnb-base": "^13.1.0",
     "eslint-plugin-import": "^2.17.2",
     "rollup": "^0.66.6",
+    "rollup-plugin-minify-html-literals": "^1.2.5",
     "rollup-plugin-node-resolve": "^3.4.0",
     "terser": "^5.2.1"
   },
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -1,18 +1,21 @@
-import resolve from 'rollup-plugin-node-resolve';
-import json from '@rollup/plugin-json';
+import resolve from "rollup-plugin-node-resolve";
+import json from "@rollup/plugin-json";
+
+import minifyHTML from 'rollup-plugin-minify-html-literals';
 
 export default {
-  input: 'src/main.js',
+  input: "src/main.js",
   output: {
-    file: 'dist/mini-graph-card-bundle.js',
-    format: 'umd',
-    name: 'MiniGraphCard',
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -21,7 +21,8 @@
   "dependencies": {
     "@kalkih/lz-string": "^1.4.5",
     "lit-element": "^2.2.1",
-    "localforage": "^1.7.3"
+    "localforage": "^1.7.3",
+    "rollup-plugin-minify-html-literals": "^1.2.5"
```
kalkih, line 25: Add as dev dependency instead.

rollup.config.js
```diff
@@ -1,18 +1,21 @@
-import resolve from 'rollup-plugin-node-resolve';
-import json from '@rollup/plugin-json';
+import resolve from "rollup-plugin-node-resolve";
```
kalkih, line 1: Please use single quotes, eslint won't error you but it's currently only configured to run the linting on the `/src` dir (I should change this to run on everything except dist).

src/main.js
```diff
@@ -179,7 +179,7 @@ class MiniGraphCard extends LitElement {
         ?labels-secondary=${config.show.labels_secondary === 'hover'}
         ?gradient=${config.color_thresholds.length > 0}
         ?hover=${config.tap_action.action !== 'none'}
-        style="font-size: ${config.font_size}px;"
+        style="font-size:${config.font_size}px;"
```
kalkih, line 182: Why all these changes in `main.js`? Makes code less readable

src/main.js
```diff
@@ -356,7 +356,7 @@ class MiniGraphCard extends LitElement {
         style="animation-delay: ${this.config.animate ? `${i * 0.5}s` : '0s'}"
         fill='none'
         stroke-dasharray=${this.length[i] || 'none'} stroke-dashoffset=${this.length[i] || 'none'}
-        stroke=${'white'}
+        stroke=#fff
```
kalkih, line 359: Need quotes here.

terser.config.json
```diff
@@ -0,0 +1,17 @@
...
+    "unsafe_arrows": true,
+    "unsafe_math": true,
+    "unsafe_undefined": true,
+    "unsafe_proto": true,
+    "unsafe_Function": true,
+    "unsafe_comps": true,
+    "hoist_funs": true,
+    "ecma": 6
+  }
+}
```
kalkih, line 17: Not sure about these options, this could potentially break the card in older browsers

package.json
Fusseldieb: Sorry! Missed that. Makes sense.

rollup.config.js
Fusseldieb: Okay! Will do.

src/main.js
Fusseldieb: Most of the changes I made were eliminating white spaces between CSS properties. Terser, even when configured aggressively, don't remove them, so I've noticed that and did it manually. Since they are just simple CSS rules, it shouldn't impact readability too much.

src/main.js
Fusseldieb: Yep, good eye. However, Chrome rendered it just fine as it seems. Also note that before `${'white'}` was also without quotes and worked just fine. Final thought: Should I include it or can we leave it out?

terser.config.json
Fusseldieb: I haven't tested it in older browsers, more tests would be needed.

src/main.js
kalkih: Include them.
`${'white'}` should be interpret as a string, hence 'white' I think.
Anyway I'm impressed it worked at all without the quotes 😄

src/main.js
kalkih: Okay, Let's keep it like this then 👍

src/main.js
Fusseldieb: The resulting string from `stroke=${'white'}` afaik are just `stroke=white`. But okay, I'll include the 'lil quotes!

kalkih: Yes, you're probably right

Fusseldieb:
@kalkih Fixed the mentioned points!

src/main.js
```diff
@@ -215,7 +215,7 @@ class MiniGraphCard extends LitElement {
     const name = this.tooltip.entity !== undefined
       ? this.computeName(this.tooltip.entity)
       : this.config.name || this.computeName(0);
-    const color = this.config.show.name_adaptive_color ? `opacity: 1; color: ${this.color};` : '';
```
maxwroc, line 218: FYI all of these changes were not necessary as all these literals will be processed by minify-html-literals lib which you have added via rollup add-on

terser.config.json
maxwroc: I think it should be fine as terser is added as the last step (after babel). But worth to double check on some older browsers/devices (like tablets - there is a lot of people using old tablets as wall-panels).

terser.config.json
```diff
@@ -0,0 +1,17 @@
+{
+  "compress": {
+    "arguments": true,
+    "booleans_as_integers": true,
+    "keep_fargs": false,
+    "module": true,
```
maxwroc, line 6: this should be used when the input is ES6 module which is not the case here

terser.config.json
```diff
@@ -0,0 +1,17 @@
...
+    "passes": 3,
+    "unsafe_arrows": true,
+    "unsafe_math": true,
+    "unsafe_undefined": true,
+    "unsafe_proto": true,
+    "unsafe_Function": true,
+    "unsafe_comps": true,
```
maxwroc, line 13: I think this won't work without `comparisons: true`. But TBH I wouldn't enable them

terser.config.json
maxwroc: Karl is right I think. These two options can break the code for devices not supporting ES6: `unsafe_arrows`, `ecma: 6`

maxwroc:
You can as well try to remove async/await from code - this can remove pretty big chunks of code.

maxwroc:
Oh wow. I was curious how big savings would it be so I have taken a look at the source code and I think it cannot work on older browsers anyway. All async functions are transpiled to generator functions (`function*` / `yield`). This means the code cannot work on older browsers already.

Fusseldieb:
At this point I'm a bit out of the loop, but thanks for reviewing further @maxwroc . If you want, you could take it from there :)

terser.config.json
OmgImAlexis: Isn't it only IE11 that doesn't support ES6?

maxwroc: No, people often use old tablets as wall panels so in theory this could break the card for them but...

as I said in a different comment - currently transpiler converts all async functions in generator functions which means the card is broken anyway as older browsers don't support them

OmgImAlexis: Might be worth looking at publishing two files that users can switch between. One for completely modern browsers and one for legacy ones?

PR #468, opened 2020-10-07
Logarithmic scale support

Hypfer:
This is #318 + the changes mentioned by @awirtz https://github.com/kalkih/mini-graph-card/pull/318#issuecomment-644567557 + docs

Without log scale:
image

With log scale:
image


leeoniya:
you'll probably want some kind of indicator that you're looking at the data on a log scale, otherwise it will almost certainly lead to mass confusion. e.g. grid lines and tick labels in https://leeoniya.github.io/uPlot/demos/log-scales.html

Hypfer:
But it is off by default so one would need to actively set logarithmic to true for that to happen

maxwroc (approved):

kalkih:
Great work 👍

PR #492, opened 2020-11-15
Make card appear in lovelace card picker

dcs8:
This will make the card appear in the Lovelace card picker (I originally made this request against master but realised I should have used Dev branch instead)

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -952,3 +952,12 @@ class MiniGraphCard extends LitElement {
 }
 
 customElements.define('mini-graph-card', MiniGraphCard);
+
+// Configure the preview in the Lovelace card picker
+window.customCards = window.customCards || [];
+window.customCards.push({
+  type: 'mini-graph-card',
+  name: 'Mini Graph Card',
+  preview: false,
+  description: 'The Mini Graph card is a minimalistic and customizable graph card',
+});


agneevX:
Did you test this successfully?

src/main.js
```diff
@@ -952,3 +952,12 @@ class MiniGraphCard extends LitElement {
...
 
 customElements.define('mini-graph-card', MiniGraphCard);
+
+// Configure the preview in the Lovelace card picker
+window.customCards = window.customCards || [];
+window.customCards.push({
+  type: 'mini-graph-card',
+  name: 'Mini Graph Card',
+  preview: false,
+  description: 'The Mini Graph card is a minimalistic and customizable graph card',
```
agneevX, line 962: Remove the comma.

dcs8: yes I tested successfully - removing the comma gives a 'missing trailing comma (comma-dangle)' error when running through npm

agneevX: From the last object, of course. https://eslint.org/docs/rules/comma-dangle

agneevX (approved):

src/main.js
dcs8: Can see you've approved now, are there any changes required?

agneevX:
```suggestion
  description: 'The Mini Graph card is a minimalistic and customizable graph card'
```

dcs8: You might have to bare with me as I'm new to js and npm, so sorry if I'm missing something. But like I said above, removing that comma gives a 'missing trailing comma (comma-dangle)' error when running through 'npm run build'. Comma dangle seems to be set to always-multiline (or possibly always). So my understanding is that no change is required.

kalkih:
Thanks for the PR, looks good 👍

PR #501, opened 2020-11-29
Enable attributes in an entity as a data source

mork2020:
Please ignore all other files then main.js
The changes in main.js allows attributes to be configured as a data source for the mini graph card.

The access to an entity value is encapulated in a new function getEntityValue. It checks wether an attribute is configured and returns the data from the attribute or from the state of the entity. Therefore all .state accesses are exchanged with the function call.

The getCache and setCache functions are adapted to save extended keys for attributes if necessary.

The fetchRecent function also needs a change, because the REST API must return all changes of the entity to access all attribute changes.

The updateEntity function is changed to filter the attributes of an entity and to access the value of an entity via getEntityValue function.

Well, using attributes from an entity could be improved using only one fetchRecent per entity, actually an entity is fetched via the REST API multiple times, for each configured attribute. 
Perhaps the caching can also be optimized, because all attributes are stored in the cache history, although not every attributes changes in each returned fetch history.


Koma-Andrea:
Nice!

RomRider:
Thanks for the PR @mork2020. However, it's not possible to ignore files in a PR :)

It seems this can be achieved using standard HA features by creating a template sensor based on an attribute of an entity. Why introduce this complexity into `mini-graph-card`?

kalkih:
Indeed, that has been my though as well, I've always pointed people towards template sensors when it comes to this particular feature request.
I understand it would be a nice addition, less hassle setting it up, but it adds quite some additional logic and complexity to the card.
Not sure if I feel like it's worth it when it already can be achieved in HA, will have to think about it.

mork2020:
My climate sensors are connected to a Homee Gateway and i access those sensors with a Node Red Script.
This scripts handles with a lot of values per sensor and creates for every Homee sensor a HA sensor,
where the values are stored in attributes. (Actually i can't guess of another implementation, 
that keeps the Node Red part as simple as it can be.).

So, those climate sensors have a lot of attributes, e.g.
batterylevel, batterylowalarm, currentvalveposition, displayorientation, feedtemperature, firmwarerevision, 
heatingmode, linkquality, malfunctionalarm, manualoperation, openwindowdetectionsensibility, softwarerevision, 
targettemperature, temperature, temperatureoffset, wakeupinterval.
Not every attribute is relevant to be used with the graph card, i use at least those attributes:
batterylevel, currentvalveposition, feedtemperature, targettemperature, temperature.

As of today i have 10 TRVs so i have to setup and maintain 50 template sensors. Those template sensors, please correct me if I am wrong,
are additionally stored in the HA recorder/database. I dont know if that has a performance impact or not. (Any thoughts?)

While testing and configuring my mini-grahp-card (with attributes), i noticed, that while adding an existing attribute to the card,
you see the already recorded history. Configuring that attribute via a template sensor means that there is "no history", 
you have to wait, until there is enough data recorded. For me thats seems only to be relevant while fiddling around via the "correct" configuration of the card, but i want to mention it.

thetic:
I'd really like to see this functionality implemented. I agree that it *can* be accomplished without direct support in the card, but it is a repetitive, error prone, and inefficient strategy. Thanks for the awesome card.

RomRider:
I've looked quickly at the content of this PR. I don't the complexity introduced by this PR is required to enable this feature.
I'll have a go at it in the next days.

kingy444:
Definitely like the look of this feature addition, hopefully it can get through - was not looking forward to templating a bunch of attributes.
Honestly cant believe this isn't standard in HA already. Why cant we access attributes directly.

RomRider:
This feature is available in [apexcharts-car](https://github.com/RomRider/apexcharts-card).
I'll add this here too so you have a choice.

RomRider:
I'm closing this, it has been released as beta through PR #564.

PR #521, opened 2020-12-28
Add median aggregate function

scop:
Median is often a more useful alternative to average.

A noteworthy side effect here is that as currently implemented, running `_median` over given items sorts them in place. If that's a problem, need to change so that we make a shallow copy of the array and operate on that.

Caveat: this patch is untested, but filing the PR here so it won't be forgotten and to avoid duplicate work in case someone else's thinking about this. And who knows, it might actually work :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,7 @@ properties of the Entity object detailed in the following table (as per `sensor.
 | name | string |  | Set a custom display name, defaults to entity's friendly_name.
 | color | string |  | Set a custom color, overrides all other color options including thresholds.
 | unit | string |  | Set a custom unit of measurement, overrides `unit` set in base config.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -194,6 +194,7 @@ These buckets are converted later to single point/bar on the graph. Aggregate fu
 | Name | Since | Description |
 |------|:-------:|-------------|
 | `avg` | v0.8.0 | Average
+| `median` | v0.9.5 | Median
```
RomRider, line 197: Could you please remove the version number and replace it with `NEXT_VERSION` instead?

RomRider:
That look great, thanks

README.md
scop: Done and rebased against current dev.

RomRider:
Fixed and merged, thanks :)

PR #533, opened 2021-01-10
Update default path to module

OndraM:
This is where javascript module is installed by default when added via HACS web GUI. As HACS is listed as the first and also - according to the docs - recommended way, it makes sense to be in pair with this default path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,15 +37,18 @@ If you configure Lovelace via YAML, add a reference to `mini-graph-card-bundle.j
 
   ```yaml
   resources:
-    - url: /local/mini-graph-card-bundle.js?v=0.10.0
+    - url: /local/community/mini-graph-card/mini-graph-card-bundle.js?v=0.10.0
       type: module
   ```
+
+If you did not install via HACS, update the url above according to the path where you put the downloaded file.
+Note that files put in `config/www/` directory are served on `/local/` URL.
 
 Else, if you prefer the graphical editor, use the menu to add the resource:
 
 1. Make sure, advanced mode is enabled in your user profile (click on your user name to get there)
 2. Navigate to Configuration -> Lovelace Dashboards -> Resources Tab. Hit orange (+) icon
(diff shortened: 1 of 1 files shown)


kalkih:
Those instructions are for an alternative installation method (**CLI install**), so adding the HACS path wouldn't make sense.

OndraM:
How so? You need to add the resource even if you install via HACS, don't you?

(At least I have to, otherwise I was getting `Custom element doesn't exist: mini-graph-card`.)

kalkih:
That's true if you're using Lovelace in YAML mode, it's however handled automatically in storage mode.
But changing the url would led to even more confusion since both the **CLI install** and **Manual install** are referring to [Add resource reference](https://github.com/kalkih/mini-graph-card#add-resource-reference).
There are no instructions for installing with HACS, just the mention of the card being available in HACS, so if anything we should expand on those instructions.

OndraM:
Well sorry then about my confusion. However I did not setup Lovelace via YAML mode but without adding the resource via "Configuration -> Lovelace Dashboards -> Resources Tab", the panel was not working.

I basically only installed it via HACS, then I got "Custom element doesn't exist: mini-graph-card" when attempting to add to Lovelace, so I found in the issues I need to add the resource. So I followed the steps in README, which states "Enter URL /local/mini-graph-card-bundle.js" - and that obviously did not work (this is what got me an idea to improve the docs).

I admit I'm a beginner with HACS addons, but this is how my onboarding experience looked like :thinking: . Should I have followed some different steps? Or it was supposed to be working without adding the resource? 

So this is just a feedback that README confused me and required searching via issues, and was not seamless. I figured it out, but it may be improved for other newcomers, imho. 

Thanks for the nice addon, though, I really like it!

kalkih:
For sure, I appreciate your feedback, but I think it would be better to elaborate on the HACS part of the instructions since the "Add resource reference" instructions are written for the two other installation methods.

Yes, HACS should add your resources automatically.
See https://github.com/hacs/integration/issues/1659 as some users seem to have the same issue as you.

PR #537, opened 2021-01-16
chore(ci): Add Devcontainer and github action for auto-release management

RomRider:
Requires: docker installed and running

Usage:
vscode --> Reopen in contrainer
in one vscode terminal:
`container start`
in another vscode terminal
`npm run watch`

In the debug panel of vscode: `Launch Chrome Against Localhost`

Enjoy :)


kalkih (approved):
Awesome!
Seems like we should bump the node version in `.nvmrc` as well as in our travis builds.
`v8.12.0` is getting quite old and doesn't seem to play nice with the updated rollup.

PR #546, opened 2021-01-20
add multiplier to be able to convert units

davydes:
Support state multiplication to an arbitrary multiplier

I want convert hPa to mmHh, for example

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Features
 
+* Support multiplication of state to be able to convert units
 * Format numbers according to selected language [[#495](https://github.com/kalkih/mini-graph-card/issues/495) [#509](https://github.com/kalkih/mini-graph-card/issues/509)] ([875b557](https://github.com/kalkih/mini-graph-card/commit/875b55743a0152e7630a6111d7128a2a28657c97))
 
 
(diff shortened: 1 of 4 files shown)


RomRider:
Thanks for the PR.

I think this should be achieved using a template sensor in home assistant.

Today you want to add multiply, tomorrow someone else will want divide, additions, subtractions or complex math formulas.
I believe we shouldn't introduce that option in the card as it is already a feature of the backend.

Thoughts @kalkih?

davydes:
> Thanks for the PR.
> 
> I think this should be achieved using a template sensor in home assistant.
> 
> Today you want to add multiply, tomorrow someone else will want divide, additions, subtractions or complex math formulas.
> I believe we shouldn't introduce that option in the card as it is already a feature of the backend.
> 
> Thoughts @kalkih?

template sensor produces a new sensor which stores history, why we have factor value but no simple multiplier?

RomRider:
`factor_value` produces the same data, but scaled. It wouldn't be the case with a `multiplier` in your use case. So from my POV this should be a separate sensor.

kalkih:
I'm think the same as @RomRider, should be done in a template sensor, where you also specify the `unit_of_measurement`, makes sense imo.

davydes:
A bit strange opinion, `factor_value` is ok but `multiplier` isn't. Let's get rid of any data transformations or let's introduce more flexible way, but ok, as you want...

PR #564, opened 2021-02-14
feat(attribute): Retrieve an attribute instead of the state

### LINKED ISSUE #245: Feature Request: Native support for climate entities
This component is great but lacks one important feature - taking a climate entity and displaying the three sensors from it - target temperature, current temperature and climate state (hvac action idle/heat/cool).

Its currently possible to define template sensors for target temperature and current temperature to use in mini-graph but not climate state. History-graph shows climate state active (heating or cooling) as shaded area under the line graph.

Native support would be perfect.

### LINKED ISSUE #411: Request: attribute variable
Few months back the [Entity Card](https://www.home-assistant.io/lovelace/entity/#attribute) had a new variable added called "attribute".  I was wonder if this could be something added to the min-graph-card at some point. Something like the following:

```
  - entity: sensor.some_json_data
    attribute: anAttrValueFromEntity
```

RomRider:
Fixes #411, Fixes #245, Fixes #501

### DIFF
--- a/.devcontainer/ui-lovelace.yaml
+++ b/.devcontainer/ui-lovelace.yaml
@@ -13,6 +13,12 @@ views:
         show:
           extrema: true
       - type: custom:mini-graph-card
+        entities:
+          - entity: weather.home
+            attribute: pressure
+        show:
+          extrema: true
+      - type: custom:mini-graph-card
         hours_to_show: 1
         points_per_hour: 60
         entities:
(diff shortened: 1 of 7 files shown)


RomRider:
Don't comment on a close PR please. If your sensor is excluded from the recorder, it will not work. If it's not working, please open a new issue.

MaxVRAM:
Thanks for the tip. Thought I shouldn't open an issue considering this feature hasn't made it to proper release yet, so was tying to keep it local to the version. Please let me know how to approach this in future. For the record, recorder is not set to exclude anything.

PR #585, opened 2021-03-26
Improve label readability on small cards

KNXBroker:
There is often the problem on small cards that the bottom label is hidden.

This PR does not completely solve the problem, as the bottom label can never be completely seen if the graph height is too small.  However, it improves the readability significantly. 

- The top and bottom label padding will be set to 0, as this padding does not have any benefit.
- The line height of the labels is set to normal ( template independent) which makes the bubbles smaller (at same font size).

image

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -323,7 +323,7 @@ const style = css`
     font-weight: 400;
     justify-content: space-between;
     margin-right: 10px;
-    padding: .6em;
+    padding: 0 .6em;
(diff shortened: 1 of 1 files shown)


KNXBroker:
Sorry, I think I have to revert this PR. Today I did an additional test: Bottom padding is required for filled graphs and top border is required for cards without state / name. Therefore this is not an solution.

Issue #182, opened 2019-10-22
Feature request: relative color thresholds


bokub:
Hello,
I have an idea to improve the **Dynamic line color** feature.

Instead of setting absolute values in the `color_thresholds` option, it would be great to use relative values.

For example:

```yaml
  color_thresholds:
    - value: 0%
      color: "#1feaea" # Blue
    - value: 50%
      color: "#ffd200" # Yellow
    - value: 100%
      color: "#f72047" # Red
```

Or even easier, omitting values 
```yaml
  color_thresholds:
    - color: "#1feaea" # Blue
    - color: "#ffd200" # Yellow
    - color: "#f72047" # Red
```

![image](https://user-images.githubusercontent.com/17952318/67288877-f9f45580-f4dd-11e9-96f3-d42115ef7dc6.png)

The goal is that the bottom of the graph is always blue and the top is always red, no matter what the values are.

----

Technically, this is easy to do in SVG: 

```svg
<defs>
    <linearGradient id="my-gradient-id" x1="0" y1="1" x2="0" y2="0">
        <stop offset="0" stop-color="#1feaea"></stop>
        <stop offset="0.5" stop-color="#ffd200"></stop>
        <stop offset="1" stop-color="#f72047"></stop>
    </linearGradient>
</defs>
<path stroke="url(#my-gradient-id)" d="..."></path>

```

kalkih:
Thanks, great suggestion!

jeepotvin:
Great suggestion. Would it be possible to use dynamic values based on other sensors as threshold? For instance, if I have a sensor that is used as a maximum value, I could use (sensor.my_sensor_max_value - 5) to change line color.

PR #596, opened 2021-04-28
Interpolate stops when computing thresholds during config building

acshef:
This is for folks (like myself) that use many colors in their graphs (see [cube helix](http://www.mrao.cam.ac.uk/~dag/CUBEHELIX/)). It's no longer necessary to specify all the stop points in color thresholds (gradients). This paves the way for two more improvements (which are not featured in this PR):
* Implicitly assuming the first and last stops are the min/max values in the graph (or some other reasonable logic), thereby making all stop points completely optional
* Allowing for named color patterns (hardcoded into the project), like `"rainbow"`, `"cubehelix"`, `"rocket"`, `"mako"`, etc. (adopting common color patterns already established in other FOSS data visualization projects)

### DIFF
--- a/src/buildConfig.js
+++ b/src/buildConfig.js
@@ -22,7 +22,10 @@ const findFirstValuedIndex = (stops, startIndex) => {
       return i;
     }
   }
-  throw new Error('Could not find right-nearest valued stop');
+  throw new Error(
+    'Error in threshold interpolation: could not find right-nearest valued stop. ' +
+    'Do the first and last thresholds have a set "value"?'
+  );
 };
 
 /**
--- a/src/buildConfig.js
+++ b/src/buildConfig.js
@@ -23,8 +23,8 @@ const findFirstValuedIndex = (stops, startIndex) => {
     }
   }
   throw new Error(
-    'Error in threshold interpolation: could not find right-nearest valued stop. ' +
-    'Do the first and last thresholds have a set "value"?'
+    'Error in threshold interpolation: could not find right-nearest valued stop. '
+    + 'Do the first and last thresholds have a set "value"?',
   );
 };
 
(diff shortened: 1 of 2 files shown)


src/buildConfig.js
```diff
@@ -8,15 +8,88 @@ import {
...
+ * @param {number} startIndex
+ * @returns {number}
+ */
+const findFirstValuedIndex = (stops, startIndex) => {
+  for (let i = startIndex, l = stops.length; i < l; i += 1) {
+    if (stops[i].value != null) {
+      return i;
+    }
+  }
+  throw new Error('Could not find right-nearest valued stop');
```
kalkih, line 23: I guess this shouldn't really happen but I think it would be good to elaborate further and indicate that this is related to the color threshold option, since these errors are presented to the user in the error card.

kalkih:
Thanks for the PR, looks good, Great feature! 
See my comment & lint errors.

Should perhaps think about breaking out threshold related functions from `buildConfig.js`, but it's fine for now I think.

acshef:
Dang, my ESLint in VSCode got turned off. I was flying blind!

acshef:
@kalkih I found a bug. Is there a way you can "re-open" the PR? If not, should I open a new PR to `master`? Or `dev`?

kalkih:
No worries, just open another PR towards `dev`.

drthanwho:
> Implicitly assuming the first and last stops are the min/max values

Would that mean first and last values that are hardcoded to be considered min/max? 
I think it could be useful to be able to assign first and last stops as min and max values for the given _hours_to_show._ Rather than having to hardcode the values

bokub:
> I think it could be useful to be able to assign first and last stops as min and max values for the given hours_to_show. Rather than having to hardcode the values

That would be great, yes! 

Related:  #182

PR #600, opened 2021-05-06
Delta range

mhspelt:
Hi, this is my first pull request ever so I hope I did everything right. 
I have a sensor in HA that can go negative during the day (it measures net power usage taking solar into account) and I wanted to know the difference between the first reading of the day and the final reading of the day. So I changed _delta_ to calculate this (for monotonously increasing data the result is identical), and renamed what _delta_ originally did to _range_. I hope you'll accept my PR; thanks!

### DIFF
--- a/src/graph.js
+++ b/src/graph.js
@@ -15,6 +15,7 @@ export default class Graph {
       last: this._last,
       sum: this._sum,
       delta: this._delta,
+      range: this._range,
     };
 
     this._history = undefined;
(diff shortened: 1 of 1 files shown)


RomRider:
Could you please PR against dev?
Also this would be a breaking change, I'd suggest you use the same as what I did in apexcharts-card to keep some consistency. There what you are trying to achieve is called `diff` and delta should remain delta.
Thanks

OmgImAlexis:
@mhspelt please rebase.

jlsjonas:
Superseded by #601

Issue #591, opened 2021-04-15
value_factor is confusing


mflagler:
Scaling values using the value_factor is too difficult. This should simply be a multiplier, not a factor/exponent of 10 that you want to scale values by.  For converting MB to GB or GB to MB it makes it much easier, and for anything that is scaled by 10s then you just multiply by the factor of 10, 100, or 0.1, or 0.01 and it's plain and simple. There should be a pre-scale offset and a post-scale offset for anyone wanting to convert temps between C and F. Then you could pretty much convert between any units anyone should ever need.

ildar170975:
In the Docs:
![image](https://user-images.githubusercontent.com/71872483/153637050-4d70642c-4907-4f6e-aa0c-7dca3d03caa0.png)

IMHO the `value_factor` should be ANY number [-xxx.xxx, +xxx.xxx].
Values like "0.9, 0.5, 0.04" should scale the state down.
Values like "1.2, 25, 400.56" should scale the state up.
Negative values should make the state negative.

Also, should be specified per-entity - https://github.com/kalkih/mini-graph-card/issues/607

jlsjonas:
We should add a _simpler_ multiplier (or similar) property and deprecate value_factor to be removed at a later date

RcRaCk2k:
Yeah, i also need that.

Having costs per kWh and a cost average for the last hour. So the line of costs per kWh is so small that you can not see any changes in the cent-values, so i have to multiply that bei 5 times, so i can see the price changes in the chart.

Danie10:
I was also looking for this as I have a forecasted solar radiation in W/m2, and it needed to be multiplied by a factor of 15.57 to equate to the same power as the area of my solar panels. Was looking to have a 'value_factor: 15.57' for that entity.

ildar170975:
check v0.14

PR #608, opened 2021-05-24
feat: add value_multiplier and extend value_factor

### LINKED ISSUE #607: value_factor should be per-entity
A global value_factor seems less useful -- it implies that all values already have the same base, at which point you could get any affect you'd like by just playing with graph ranges.  It would be very handy if a multiplier (see #591) could be done per entity to graph.  A simple use case would be graphing a percentage charge (0-100 value) on the same graph as instantaneous kW draw (~0-10 or so values).

OmgImAlexis:
This adds value_multiplier which similarly to the value_factor only it multiplies the value directly. By default it's set to `1`.

I've also added the `value_factor` to the entity itself to allow different factors for each entity. Personally I like this as it allows cards to show more than a single type of data. For example I want to show data usage and power usage on one, the issue is one is in bytes(I want MB) and one is watts(I want KW). I need one to be divided by `1024` and the other `1000`.

I haven't tested this yet so please let me know if there's anything I need to fix.

Closes https://github.com/kalkih/mini-graph-card/issues/607 and https://github.com/kalkih/mini-graph-card/issues/591

### DIFF
--- a/src/buildConfig.js
+++ b/src/buildConfig.js
@@ -52,6 +52,7 @@ export default (config) => {
     state_map: [],
     cache: true,
     value_factor: 0,
+    value_multipler: 1,
     tap_action: {
       action: 'more-info',
     },
--- a/src/graph.js
+++ b/src/graph.js
@@ -5,7 +5,19 @@ import {
 } from './const';
 
 export default class Graph {
-  constructor(width, height, margin, hours = 24, points = 1, aggregateFuncName = 'avg', groupBy = 'interval', smoothing = true, logarithmic = false) {
+  constructor({
+    width,
+    height,
+    margin,
+    hours = 24,
+    points = 1,
+    aggregateFuncName = 'avg',
+    valueFactor = 0,
+    valueMultiplier = 1,
+    groupBy = 'interval',
+    smoothing = true,
+    logarithmic = false
+  }) {
     const aggregateFuncMap = {
       avg: this._average,
       max: this._maximum,
@@ -26,6 +38,8 @@ export default class Graph {
     this.points = points;
     this.hours = hours;
     this.aggregateFuncName = aggregateFuncName;
+    this.valueFactor = valueFactor;
+    this.valueMultiplier = valueMultiplier;
     this._calcPoint = aggregateFuncMap[aggregateFuncName] || this._average;
     this._smoothing = smoothing;
     this._logarithmic = logarithmic;
--- a/src/main.js
+++ b/src/main.js
@@ -100,25 +100,26 @@ class MiniGraphCard extends LitElement {
     const entitiesChanged = !compareArray(this.config.entities || [], config.entities);
 
     this.config = buildConfig(config, this.config);
-
     if (!this.Graph || entitiesChanged) {
       if (this._hass) this.hass = this._hass;
       this.Graph = this.config.entities.map(
-        entity => new Graph(
-          500,
-          this.config.height,
-          [this.config.show.fill ? 0 : this.config.line_width, this.config.line_width],
-          this.config.hours_to_show,
-          this.config.points_per_hour,
-          entity.aggregate_func || this.config.aggregate_func,
-          this.config.group_by,
-          getFirstDefinedItem(
+        entity => new Graph({
+          width: 500,
+          height: this.config.height,
+          margin: [this.config.show.fill ? 0 : this.config.line_width, this.config.line_width],
+          hours: this.config.hours_to_show,
+          points: this.config.points_per_hour,
+          aggregateFuncName: entity.aggregate_func || this.config.aggregate_func,
+          valueFactor: entity.valueFactor || this.config.valueFactor,
+          valueMultiplier: entity.valueMultiplier || this.config.valueMultiplier,
+          groupBy: this.config.group_by,
+          smoothing: getFirstDefinedItem(
             entity.smoothing,
             this.config.smoothing,
             !entity.entity.startsWith('binary_sensor.'), // turn off for binary sensor by default
           ),
-          this.config.logarithmic,
-        ),
+          logarithmic: this.config.logarithmic,
+        }),
       );
     }
   }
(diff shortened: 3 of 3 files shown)


maulwuff:
Is there anything missing to complete this PR? 
I'd like to compare sun elevation with solar power and I need to scale the evelation up to make it visible :)

OmgImAlexis:
@maulwuff I don't think so.

kalkih:
> Is there anything missing to complete this PR?
> I'd like to compare sun elevation with solar power and I need to scale the evelation up to make it visible :)

The review comments need to be addressed before we can merge this.

OmgImAlexis:
> The review comments need to be addressed before we can merge this.

There are no review comments..?

kalkih (changes requested):

maulwuff:
@OmgImAlexis come on, you can fix that ;-)

smiba:
Would love to see this in mini-graph-card, currently this prohibits me from making a total power draw card. My total consumption is reported in kW, while individual devices report in W.

kalkih:
> Would love to see this in mini-graph-card, currently this prohibits me from making a total power draw card. My total consumption is reported in kW, while individual devices report in W.

In the meantime you could look at creating a template_sensor and handle the conversion there.

jlsjonas:
@OmgImAlexis Do you need some help addressing these (minor) comments & getting this merged?

ildar170975:
Please check this: https://github.com/kalkih/mini-graph-card/issues/591#issuecomment-1036431648
As far as I understood, this PR introduces a new `value_multiplier` property in addition to the already present (but not per-entity) `value_factor` property.
IMHO there should be ONE multiplying property.

andrewjswan:
Great idea, then I can invert the graph to get this result:
image

ildar170975:
No feedback from the author for a long time...
Closed in favour of https://github.com/kalkih/mini-graph-card/pull/1359 which supports both "exponent" factor & proposed here "multiplier" factor.

andrewjswan:
> Closed in favour of #1359 which supports both "exponent" factor & proposed here "multiplier" factor.

But it still doesn't allow you to invert the graph as shown in the example https://github.com/kalkih/mini-graph-card/pull/608#issuecomment-1129812319

ildar170975:
> But it still doesn't allow you to invert the graph

1. Although this PR proposed an individual per-entity factor, it was incomplete. Adding that per-entity factor will be a next improvement after https://github.com/kalkih/mini-graph-card/pull/1359, it is planned.
2. A negative factor is not a sufficient way to invert a graph since it will give negative value/labels/extremas. A corresponding "invert /mirror graph" feature is planned.

andrewjswan:
> 2\. A negative factor is not a sufficient way to invert a graph since it will give negative value/labels/extremas.

In my opinion, this doesn't matter, but it does highlight a problem with the architecture—in theory, all data coefficients should affect only the graph, not the values. But overall, any solution will work.

ildar170975:
@andrewjswan 
The `invert` option is one of the 1st priority.
Надо чуть подождать))

Meanwhile, to display an "inverted" graph you can use card-mod - https://community.home-assistant.io/t/lovelace-mini-graph-card/71385/1936?u=ildar_gabdullin

image

More card-mod styles here -  https://community.home-assistant.io/t/card-mod-add-css-styles-to-any-lovelace-card/120744/1396

ildar170975:
Proof of concept for an inverted graph, coming soon:

image

andrewjswan:
> Proof of concept for an inverted graph, coming soon:

In my opinion, inverting along the second axis is the worst option. If you look at the traffic graph, In/Out is on one axis, Out is inverted, and the second axis can contain other data.

andrewjswan:
Screenshot_2026-09-01-08-23-30-767_io homeassistant companion android-edit

ildar170975:
As was already told, ypu cannot invert a graph w/o inverting a scale; applying a negative factor is a bad way since it will show negative values.
As for "secondary axis is used for other needs" - come on, you need it for a background graph like "day/night", "on/off" - and all these binary-based graphs will be a separate functionality with VERTICAL transitions (will not need a dedicated axis).

andrewjswan:
I don't see any problems. Here's a chart that provides a great understanding of the scale and trends in traffic. It uses two axes: the first shows traffic, and the second shows additional data. If you were to invert the axes, you'd need to add at least one more axis.

Screenshot_2026-09-01-08-36-24-626_io homeassistant companion android-edit




> As for "secondary axis is used for other needs" - come on, you need it for a background graph like "day/night", "on/off" - and all these binary-based graphs

Not just the amount of free storage space, remaining data allowance under your plan, and so on.

> since it will show negative values.

Oh, come on—there aren't any negative values on the graph above, and everything is perfectly clear.

ildar170975:
Assume you applied a negative factor to a graph.
Then you will see a nice line - but wrong states if you switch them on (and a negative Y scale if you switch labels on).
(not to mention possible future extrema/avg data lines supported for other entities)
The fact that you do not see negative states does not mean they are not.
Means - applying a negative factor is a mere hack which leads to an inconsistency.

Factor (exponential, scale) - are only supposed to be used when you do not want to create template sensors with a required conversion; using a negative -1 factor only to see an inverted graph is a hack which hopefully will be not needed in future.

andrewjswan:
> The fact that you do not see negative states does not mean they are not.
> Means - applying a negative factor is a mere hack which leads to an inconsistency.

Inconsistent with what? Let’s say there are two metrics on the graph—Download and Upload. I apply a factor of -1 to Upload and get an inverted graph. Even if I see -20 Mb there, I fully understand that I sent 20 Mb. And there are no inconsistencies. And everything is perfectly clear, both in terms of the values and the axes. If I don’t need this behavior, I simply don’t apply the -1 multiplier and can use your inversion method instead—but then I lose the functionality of the second axis.
Everyone will choose the approach that works best for them. That’s the beauty of customization.

> Factor (exponential, scale) - are only supposed to be used when you do not want to create template sensors with a required conversion; using a negative -1 factor only to see an inverted graph is a hack which hopefully will be not needed in future.

Well, since we're expanding the customization options, why should we restrict people in any way? If someone sets the coefficient to -1, they understand the consequences and are comfortable with them.

ildar170975:
The inconsistency comes when you show a state of the “minused” entity - it will show a negative state. Repeat, your “-1” method is a hack with a limited application “do not switch on showing a state, do not switch on showing axis labels”.
As long as you agree to use this hack - you are welcome, but we should provide/use consistent methods. 
Also, I agree that the secondary axis becomes occupied in case of the “my” method. Thinking about a tertiary axis without showing axis labels.

andrewjswan:
Inversion around 0 is simply the transformation of 5 to -5. It doesn’t matter how it’s done—whether by a flag or a coefficient. And since the user has chosen inversion, they understand that the values on the axes may change. I don’t see any problem with displaying values and axes that will have negative values for the inverted graph, since anything less than 0 is negative.

PR #626, opened 2021-06-30
Update README.md

flyize:
Update resource paths.


kalkih:
Hello,
These resource paths are provided for manual & cli installation methods, they should be correct if you follow the manual or & CLI installation instructions.

PR #658, opened 2021-10-13
Fix #269: add order attribute to entities

### LINKED ISSUE #269: Graph lines should be drawn in reverse order
When drawing my graph three series on it the last series specified gets drawn on top of the others...

I think that this should be the other way around ... so the "primary" appears on top.

ricardoboss:
This PR adds an optional attribute to entities, which changes the order, in which graphs are rendered. This fixes #269: if no order is set, the graphs are rendered in reversed order as they are specified in the config. This means the first entity is rendered last (on top of everything else).

Nothing else should be affected by this. Animations should still be played from back to front etc.

Please tell me what you think and test this enhancement with your own setup!

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -515,10 +515,10 @@ class MiniGraphCard extends LitElement {
 
     // map element keys to render function
     const fillMap = orderedKeys.map((i, order) => this.renderSvgFill(this.fill[i], i, order));
-    const fillRectMap = orderedKeys.map((i) => this.renderSvgFillRect(this.fill[i], i));
+    const fillRectMap = orderedKeys.map(i => this.renderSvgFillRect(this.fill[i], i));
     const lineMap = orderedKeys.map((i, order) => this.renderSvgLine(this.line[i], i, order));
-    const lineRectMap = orderedKeys.map((i) => this.renderSvgLineRect(this.line[i], i));
-    const barMap = orderedKeys.map((i) => this.renderSvgBars(this.bar[i], i));
+    const lineRectMap = orderedKeys.map(i => this.renderSvgLineRect(this.line[i], i));
+    const barMap = orderedKeys.map(i => this.renderSvgBars(this.bar[i], i));
     const pointsMap = orderedKeys.map((i, order) => this.renderSvgPoints(this.points[i], i, order));
 
     return svg`
--- a/src/main.js
+++ b/src/main.js
@@ -109,11 +109,12 @@ class MiniGraphCard extends LitElement {
       if (this._hass) this.hass = this._hass;
       let order = 0;
       this.Graph = this.config.entities.map(
-        entity => {
+        (entity) => {
           let graphOrder = entity.order;
-          if (!Number.isInteger(graphOrder))
-            graphOrder = order++;
-          else
+          if (!Number.isInteger(graphOrder)) {
+            order += 1;
+            graphOrder = order;
+          } else
             order = graphOrder + 1;
 
           return new Graph(
@@ -132,7 +133,7 @@ class MiniGraphCard extends LitElement {
             ),
             this.config.logarithmic,
           );
-        }
+        },
       );
     }
   }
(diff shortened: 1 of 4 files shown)


jlsjonas:
@ricardoboss could you please resolve the (CI) issues reported for your PR so we can review & include it, thanks!

ricardoboss:
Will do in the near future!

ricardoboss:
@jlsjonas waiting for approval

jlsjonas:
@ricardoboss looks like there are still some linting issues :/

ricardoboss:
@jlsjonas fixed

jlsjonas:
@ricardoboss close... 😅

ricardoboss:
Damn, they were last-minute changes. Didn't notice linting fails... Now we should be good! @jlsjonas

src/main.js
```diff
@@ -496,20 +507,34 @@ class MiniGraphCard extends LitElement {
...
-          ${this.line.map((line, i) => this.renderSvgLineRect(line, i))}
-          ${this.bar.map((bars, i) => this.renderSvgBars(bars, i))}
+          ${fillMap}
+          ${fillRectMap}
+          ${lineMap}
+          ${lineRectMap}
+          ${barMap}
         </g>
-        ${this.points.map((points, i) => this.renderSvgPoints(points, i))}
+        ${pointsMap}
```
jlsjonas, line 537: While we might want to rework this, not ever map is based on `this.fill` (`show_fill` can be set to false); there's also no real benefit moving these to `const`'s as it's the only usage, and the function is returning svg elements

jlsjonas (changes requested):
I feel like the order attribute is adding considerable complexity to the code for a limited benefit, vs. just reversing the drawing order to have the primary (first listed) entity drawn on top.

Can you provide a good use-case for specifying an order vs. just reversing the drawing direction (so that the primary entity ends up on top)?

ricardoboss:
Well I want my graphs ordered in a strict way without having to reorder the entries in yml. If this isn't reason enough then I can only refer to the original issue, where people explicitly wished for that feature too.

src/main.js
ricardoboss: Honestly, I made these changes in about two days and I can only remember pieces of the code. I'm not in the mindset for deep discussions about details right now. If you have a better solution I'm happy to adjust my code!

arigit:
@jlsjonas the first entry in yml is the one that uses the primary accent color of the theme. Manually changing the order triggers a lot of manual work in the graph and also makes the graphs look weird color-wise when you change the theme (i.e. you need to readjust). If different users use different themes, then graphs can become unreadable (if one is a dark theme and the other one is light)

Bottom line, we need this.
Thanks!

ricardoboss:
Look, it works for me and I am not willing to work on this anymore (I can only remember having to dive deep into the code and being frustrated). If anyone wants to pick up where I left, feel free. I'll leave my branch as it is.

That said, I will close this PR then.

jlsjonas:
> the first entry in yml is the one that uses the primary accent color of the theme.

@arigit Yes, and I fully agree that we need to have the first entry (using the primary accent color) _on top_, instead of on the bottom as it currently is.

> Manually changing the order triggers a lot of manual work in the graph and also makes the graphs look weird color-wise when you change the theme (i.e. you need to readjust).

That's exactly what I'm worried about... this PR is about adding an `order` entry to the elements; intuitively, the order is implied by the order of entities. This would break away from that, possibly leading to confusing & definitely leading to more complexity maintaining the code & adding new functionality


> it works for me and I am not willing to work on this anymore (I can only remember having to dive deep into the code and being frustrated).

Fully understandable, my first deep-dive into the codebase was quite frustrating too; this is something I'd take up under the v0.12 release regardless (thank you for your contribution regardless, and for leaving the branch available).
I'm, to be clear, not saying we absolutely can't have an order attribute, but I want to make sure that it would be _really_ something we'd need. As we have limited (spare time) resources maintaining this card, we have to ensure that new functionality is important enough to most users, to avoid having the project becoming too complex to maintain in the future.

My apologies that this only came so late into the process, we're still actively reviving the project & going over older issues.

For others stumbling upon this, let's continue the discussion (if needed) in #269

src/main.js
jlsjonas: This comment was not meant as a deep discussion; the proposed change would skip entities where `show_fill` is set to false. But let's park this until we have a decision regarding the `order` attribute in general (and again, thanks for your contribution regardless)

PR #677, opened 2021-12-25
Show last point value as state

jlsjonas:
Add support for displaying the last (point) value as state (particularly useful for `delta` aggregation)

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ All properties are optional.
 |------|:-------:|:-------:|-------------|
 | name | `true` | `true` / `false` | Display name.
 | icon | `true` | `true` / `false` | Display icon.
-| state | `true` | `true` / `false` / `'last'` | Display current state. `'last'` will show the last graph point's value.
+| state | `true` | `true` / `false` / `last` | Display current state. `last` will show the last graph point's value.
 | graph | `line` | `line` / `bar` / `false` | Display option for the graph. If set to `bar` a maximum of `96` bars will be displayed.
 | fill | `true` | `true` / `false` / `fade` | Display the line graph fill.
 | points | `hover` | `true` / `false` / `hover` | Display graph data points.
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ All properties are optional.
 |------|:-------:|:-------:|-------------|
 | name | `true` | `true` / `false` | Display name.
 | icon | `true` | `true` / `false` | Display icon.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -153,7 +153,7 @@ All properties are optional.
 |------|:-------:|:-------:|-------------|
 | name | `true` | `true` / `false` | Display name.
 | icon | `true` | `true` / `false` | Display icon.
-| state | `true` | `true` / `false` | Display current state.
+| state | `true` | `true` / `false` / `'last'` | Display current state. `'last'` will show the last graph point's value.
```
kalkih, line 156: Remove double quotes around `last` here.

src/main.js
```diff
@@ -253,11 +253,13 @@ class MiniGraphCard extends LitElement {
...
-        ? this.entity[0].attributes[this.config.entities[0].attribute]
-        : this.entity[0].state
-    );
+    const state = value !== undefined ? value : (this.config.show.state === 'last'
+      ? this.points[0][this.points[0].length - 1][V]
+      : (
+        this.config.entities[0].attribute
+          ? this.entity[0].attributes[this.config.entities[0].attribute]
+          : this.entity[0].state
+      ));
```
kalkih, line 262: This is getting quite complex to read.
I would prefer if we could avoid these nested tenary operators by replacing with simpler `if` / `else` statements or similar.

src/main.js
```diff
@@ -764,6 +766,7 @@ class MiniGraphCard extends LitElement {
             && config.entities[i].show_fill !== false) this.fill[i] = this.Graph[i].getFill(line);
           if (config.show.points && (config.entities[i].show_points !== false)) {
             this.points[i] = this.Graph[i].getPoints();
+            console.log(`### lastpoint ${this.points[i][this.points[i].length - 1][V]}`);
```
kalkih, line 769: Remove comment.

kalkih (changes requested):

src/main.js
jlsjonas: Whoops :)

src/main.js
jlsjonas: Considering a refactor _soon_ to allow most (sensible) configuration items (like value_factor) to be configurable both on root level (for all entities) or on a specific entity; which should also improve readability of this blob (converted to if/else statements _for now_ )

kalkih (approved):

PR #691, opened 2022-01-15
Unify state rendering to avoid discrepancies

### LINKED ISSUE #580: Attribute feature in v0.11.0-dev.3 only works for the first entity
Does the attributes function in this branch work? I'm not having luck while using the following lovelace code:

```
type: 'custom:mini-graph-card'
entities:
  - entity: switch.hs110_server
    name: Watts
    attribute: current_power_w
```

Developer tools for the switch.hs110_server entity displays the following:

```
current_power_w: '87.49'
total_energy_kwh: '18.060'
voltage: '247.2'
current_a: '0.47'
today_energy_kwh: '1.844'
friendly_name: HS110 Server
```

My mini-graph-card display N/A for the switch.hs110_server entity and current_power_w attribute.

I have not excluded anything from recorder, and the attribute is confirmed to be listed in HA. Am I missing something here?

jlsjonas:
Fixes #580 & groundwork for entity-level configuration alignment ([!3](https://github.com/kalkih/mini-graph-card/milestone/3))

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -257,7 +257,7 @@ class MiniGraphCard extends LitElement {
       return html`
         <div class="states flex" loc=${this.config.align_state}>
           ${this.renderState(firstEntityConfig, 0)}
-          <div class="states--secondary">${this.config.entities.map((ent, i) => i > 0 && this.renderState(ent, i) || '')}</div>
+          <div class="states--secondary">${this.config.entities.map((entityConfig, i) => i > 0 && this.renderState(entityConfig, i) || '')}</div>
           ${this.config.align_icon === 'state' ? this.renderIcon() : ''}
         </div>
       `;
--- a/src/main.js
+++ b/src/main.js
@@ -252,55 +252,46 @@ class MiniGraphCard extends LitElement {
   }
 
   renderStates() {
-    const { entity, value: tooltipValue } = this.tooltip;
-    let state;
     const [firstEntityConfig] = this.config.entities;
-    if (tooltipValue !== undefined) {
-      state = tooltipValue;
-    } else if (this.config.show.state === 'last') {
-      state = this.points[0][this.points[0].length - 1][V];
-    } else {
-      const [firstEntity] = this.entity;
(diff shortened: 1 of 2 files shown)


src/main.js
```diff
@@ -252,55 +252,46 @@ class MiniGraphCard extends LitElement {
...
-              ${this.computeState(state)}
-            </span>
-            <span class="state__uom ellipsis" style=${color}>
-              ${this.computeUom(entity || 0)}
-            </span>
-            ${this.renderStateTime()}
-          </div>
-          <div class="states--secondary">${this.config.entities.map((ent, i) => this.renderState(ent, i))}</div>
+          ${this.renderState(firstEntityConfig, 0)}
+          <div class="states--secondary">${this.config.entities.map((ent, i) => i > 0 && this.renderState(ent, i) || '')}</div>
```
kalkih, line 260: I know this was already poorly named before but I would suggest

```suggestion
          <div class="states--secondary">${this.config.entities.map((entityConfig, i) => i > 0 && this.renderState(entityConfig, i) || '')}</div>
```

src/main.js
```diff
@@ -716,13 +707,8 @@ class MiniGraphCard extends LitElement {
   }
 
   numberFormat(num, language, dec) {
-    if (!Number.isNaN(Number(num)) && Intl) {
-      if (dec === undefined || Number.isNaN(dec)) {
```
kalkih, line 720: I don't remember exactly why I added this guard, but I suspect there was issues if `dec` was `undefined` or not a number when used in `Intl.NumberFormat()` below.

Would be great if you could double check this.

jlsjonas: I double-checked it before committing, it was added in https://github.com/kalkih/mini-graph-card/commit/875b55743a0152e7630a6111d7128a2a28657c97 (not by you :) )
if it's undefined, it works as intended (undefined is the default after all), if it's NaN it's already failing earlier (thus no functional difference, and config should be considered invalid in that case regardless => should be checked at that level with the upcoming milestone)

PR #749, opened 2022-03-03
Capitalize hacs name

TheZoker:


jlsjonas:
Hi, thanks for your contribution!
While I understand the rationale behind it, I believe that as there's no true convention for the naming style, this change would only make sense once we have a proper configuration UI; as people are already familiar with the name as-is.

Feel free to provide other contributions, ideally based on open issues, so you know that they're more likely to be accepted.

We'll reconsider this name change when you don't need to know yaml for basic operations.

PR #871, opened 2022-11-16
Add unit_first option

caraar12345:
I'm trying to use `mini-graph-card` to track energy costs but the unit shows after the monetary value. Thus showing as `2.50 £` which isn't right!

This adds a Boolean option, `unit_first`, which moves the unit to the start.

---

Fully understand that this is ... messy code; I'm not a JS dev by any means 😅 
However, just wanted to contribute my "fork" back upstream 🚀

### DIFF
--- a/.github/workflows/semantic-pull-request.yml
+++ b/.github/workflows/semantic-pull-request.yml
@@ -1,5 +1,5 @@
 name: "Semantic PR"
-description: Ensure your PR title matches the Conventional Commits spec (https://www.conventionalcommits.org/).
+# description: Ensure your PR title matches the Conventional Commits spec (https://www.conventionalcommits.org/).
 
 on:
   pull_request_target:
--- a/src/main.js
+++ b/src/main.js
@@ -570,7 +570,7 @@ class MiniGraphCard extends LitElement {
           <div class="info__item">
             <span class="info__item__type">${entry.type}</span>
             <span class="info__item__value">
-              ${this.computeState(entry.state)} ${this.computeUom(0)}
+              ${this.config.unit_first ? this.computeUom(0) this.computeState(entry.state) : this.computeState(entry.state) this.computeUom(0)}
             </span>
             <span class="info__item__time">
               ${entry.type !== 'avg' ? getTime(new Date(entry.last_changed), this.config.format, this._hass.language) : ''}
--- a/src/main.js
+++ b/src/main.js
@@ -306,8 +306,8 @@ class MiniGraphCard extends LitElement {
           ${entityConfig.show_indicator ? this.renderIndicator(state, id) : ''}
           ${this.unitFirstOrLast(isPrimary, tooltipValue, state, id, entity)}
           ${isPrimary && this.renderStateTime() || ''}
-        </div >
-        `;
+        </div>
+      `;
     }
   }
 
--- a/src/main.js
+++ b/src/main.js
@@ -273,6 +273,26 @@ class MiniGraphCard extends LitElement {
     }
   }
 
+  unitFirstOrLast(isPrimary, tooltipValue, state, id, entity) {
+    if (this.config.unit_first) {
+      return html`
+        <span class="state__uom ellipsis">
+          ${this.computeUom(isPrimary && entity || id)}
+        </span>
+        <span class="state__value ellipsis">
+          ${this.computeState(isPrimary && tooltipValue || state)}
+        </span>`;
+    } else {
+      return html`
+        <span class="state__value ellipsis">
+          ${this.computeState(isPrimary && tooltipValue || state)}
+        </span>
+        <span class="state__uom ellipsis">
+          ${this.computeUom(isPrimary && entity || id)}
+        </span>`;
+    }
+  }
+
   renderState(entityConfig, id) {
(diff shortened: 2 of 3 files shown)


src/main.js
```diff
@@ -284,15 +304,10 @@ class MiniGraphCard extends LitElement {
           @click=${e => this.handlePopup(e, this.entity[id])}
           style=${entityConfig.state_adaptive_color ? `color: ${this.computeColor(state, id)};` : ''}>
           ${entityConfig.show_indicator ? this.renderIndicator(state, id) : ''}
-          <span class="state__value ellipsis">
-            ${this.computeState(isPrimary && tooltipValue || state)}
-          </span>
-          <span class="state__uom ellipsis">
-            ${this.computeUom(isPrimary && entity || id)}
-          </span>
+          ${this.unitFirstOrLast(isPrimary, tooltipValue, state, id, entity)}
```
akloeckner, line 307: in case you like to have a look at nested templates, maybe something along these lines might work:
```js
unitFirstOrLast(state, unit) {
  if (this.config.unit_first) {
    return unit+state;
  } else {
    return state+unit;
  }
}
```

To be used as:
```
${ this.unitFirstOrLast(
  `state template`, 
  `unit template`
) }
```

This would avoid to repeat the span templates twice. But readability might suffer. I'm not sure.

akloeckner (commented):
To me, this does seem very tidy, in fact! I would prefer to not repeat the two `span`s twice. But I found no particularly readable way to this. Except nested templates maybe. But I'm no expert and "done is better than perfect". ;-)

I just have the few concrete suggestions following...

src/main.js
```diff
@@ -284,15 +304,10 @@ class MiniGraphCard extends LitElement {
...
           ${isPrimary && this.renderStateTime() || ''}
-        </div>
-      `;
+        </div >
+        `;
```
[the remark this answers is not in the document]
caraar12345: Frankly not a clue; it was either my editor being 'helpful', or me just slamming the keyboard in the hopes that it'd work 😂

akloeckner (approved):
Not sure this counts for anything. But I'd approve of this PR! :+1:

src/main.js
caraar12345: I'll take a proper look at this in the morning but I like the sound of that. Just for clarity: you mean to call `unitFirstOrLast` with the actual template `span` strings as the parameters? 

`unitFirstOrLast("<span>state</span>", "<span>unit</span>)`

I'm not actually sure that would impede readability; it'd keep the actual "style code" in the right place and just use the function to choose the order - so would probably improve it 😄

caraar12345:
Cheers @akloeckner! Really appreciated - and thanks for taking the time to give the PR such an in depth review 🚀

src/main.js
akloeckner: Sorry, I just saw your comment in my mailbox. But, yes, that's what I meant!

ildar170975:
Placing UoM before a state may be done by CSS property too, see this [example in Community](https://community.home-assistant.io/t/lovelace-mini-graph-card/71385/2185?u=ildar_gabdullin):
```
type: custom:mini-graph-card
entities:
    unit: €
icon: mdi:currency-eur
card_mod:
  style: |
    .states .state {
      flex-direction: row-reverse;
    }
```
изображение

.github/workflows/semantic-pull-request.yml
```diff
@@ -1,5 +1,5 @@
 name: "Semantic PR"
-description: Ensure your PR title matches the Conventional Commits spec (https://www.conventionalcommits.org/).
+# description: Ensure your PR title matches the Conventional Commits spec (https://www.conventionalcommits.org/).
```
jlsjonas, line 2: Why is this commented out? 🤔

src/main.js
```diff
@@ -284,12 +304,7 @@ class MiniGraphCard extends LitElement {
           @click=${e => this.handlePopup(e, this.entity[id])}
           style=${entityConfig.state_adaptive_color ? `color: ${this.computeColor(state, id)};` : ''}>
           ${entityConfig.show_indicator ? this.renderIndicator(state, id) : ''}
-          <span class="state__value ellipsis">
-            ${this.computeState(isPrimary && tooltipValue || state)}
-          </span>
-          <span class="state__uom ellipsis">
-            ${this.computeUom(isPrimary && entity || id)}
-          </span>
+          ${this.unitFirstOrLast(isPrimary, tooltipValue, state, id, entity)}
```
jlsjonas, line 307: Please use the css-based approach suggested by @ildar170975  (in the card-mod) instead, to reduce code complexity (improving maintainability)

Thanks for the contribution!

jlsjonas (changes requested):

.github/workflows/semantic-pull-request.yml
caraar12345: With it present, the GitHub action refused to run; it didn't recognise the description field!

akloeckner:
@caraar12345, I'll close this. See https://github.com/kalkih/mini-graph-card/pull/871#discussion_r1056708972 and #937 . Just re-open if you still think we should invest in this feature. 

In any case: Thank you for your efforts!

PR #877, opened 2022-12-01
feat: add scale_factor and extend value_factor (next)

X-Ryl669:
I've bootstrapped from PR #608 that was stalling. 

I've fixed the remarks and fixed the implementation to also work with negative scale factor.
This allow to have a negative `value_multiplier` (for example, when injecting current in the grid) per entity and also a `value_factor` per entity.

I've also fixed the documentation.

This gives this:
image

where the blue graph is a negative valued since it's a production not a consumption:
image

### DIFF
--- a/.devcontainer/configuration.yaml
+++ b/.devcontainer/configuration.yaml
@@ -5,6 +5,12 @@ lovelace:
       type: module
 
 default_config:
+http: # support devcontainers/codespaces/gitpod
+  use_x_forwarded_for: true
+  trusted_proxies:
+    - 127.0.0.1
+  ip_ban_enabled: true
+  login_attempts_threshold: 5
 
 demo:
 
@@ -37,4 +43,8 @@ sensor:
   - platform: random
     name: random_0_1000
     minimum: 0
+    maximum: 1000
+  - platform: random
+    name: random_150_1000
+    minimum: 150
     maximum: 1000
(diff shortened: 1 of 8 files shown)


akloeckner:
Thanks for picking this up! From first sight and without testing, I'd suggest a few changes:
* PRs should target the dev branch as far as I understood. So, maybe rebase it to get it merged more easily...
* I believe, all computations can be delegated to `computeState`, by passing entity's `index` and using defaults (`||`). See `computeUom` for example. Right now, I believe, extrema are even scaled twice, i.e., when they are updated in `updateExtrema` *and* when they are displayed using `computeState`.
* Take care to use the right precedence: entity `||` card `||` default. Right now, it looks to me as if card settings would take precedence over entity settings in `computeState`.
* There are some typos resulting in `multipler` instead of `multiplier`.

X-Ryl669:
1. The dev branch is lagging behind the master branch. Is it expected ?
2. That's what I did at first, but it doesn't work, since `computeState` is also called for non entity (like bounds). I could have broken this to 2 different functions but I think it's cleaner this way. Extrema aren't scaled twice, I'm just saving the value multiplier used when computing the extrema in the `abs` array, so I can compute their state with the appropriate multiplier when required.
3. You're right, I'll fix this.
4. I had a doubt about this from the initial PR which used `multipler`. TBH, I don't like the name "value multiplier" and I think "scale factor" would be better, what do you think ?

ildar170975:
@X-Ryl669 
Please consider correcting docs related to your changes.
The current description of the `value_factor` may not be right.

PR #880, opened 2022-12-03
fix: allow zero tooltip

### LINKED ISSUE #805: If a point with a zero value is selected - then a wrong state is displayed
My weather station measures the amount of rain with a rain counter, which sums up the continuous amount of rain. 
For showing the daily rain, I've configured the following card:
```
type: custom:mini-graph-card
name: Daily Rain
entities:
  - entity: sensor.weather_station_rain_counter
show:
  graph: bar
hours_to_show: 168
aggregate_func: delta
group_by: date
hour24: true
line_color: '#15317E'
labels: true 
```
The result seems correct:
mgc 01
27.4 mm is the total amount so far.

When hovering over the particular days, the card shows the correct value unless on that day there was no rain. Than it shows the total amount. However, IMHO it should show "0":
mcg06

Is my configuration wrong, or is it a card issue?

Thanks!

akloeckner:
fixes #805 as outlined in my comment there

Note: this could as well be fixed with ES2020's nullish coalescing operator `??`. But the source did not compile with it.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -293,7 +293,7 @@ class MiniGraphCard extends LitElement {
           style=${entityConfig.state_adaptive_color ? `color: ${this.computeColor(state, id)};` : ''}>
           ${entityConfig.show_indicator ? this.renderIndicator(state, id) : ''}
           <span class="state__value ellipsis">
-            ${this.computeState(isPrimary && tooltipValue || state)}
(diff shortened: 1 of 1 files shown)


akloeckner:
Looking a bit more into the (better) fix using ES2020 makes me think, this would require a major polishing of the used node modules. ES2020 requires an updated eslint, but that is not compatible with the used airbnb base configuration. So, I would not go this route at the current point in time.

Issue #878, opened 2022-12-01
Strange binning when using group_by in line graph


akloeckner:
I experience some strange behavior when binning data through `group_by` in a `line` graph.

Here's a working-as-expected card. Note how the rightmost point (currently hovered above) is today (2nd Dec):
![grafik](https://user-images.githubusercontent.com/7125061/205181500-882b6b11-86f5-4357-b1f8-380b7ca820da.png)

```yaml
type: custom:mini-graph-card
name: Außentemperatur
entities:
  - entity: weather.forecast_home
    attribute: temperature
    name: mean
    unit: °C
    aggregate_func: avg
hours_to_show: 72
group_by: date
hour24: true
decimals: 1
smoothing: false
show:
  graph: line
  labels: true
```

If I set the `hour_to_show` to something not-a-multiple-of-24, say `73` I get the following card. Note how the rightmost point is yesterday (1st Dec). It also seems shifted to the left a bit.
![grafik](https://user-images.githubusercontent.com/7125061/205181686-3d267ad3-794d-42dd-a83c-ca6876a186b5.png)

This issue does not happen with a `bar` graph.

ildar170975:
@akloeckner 
The issue is associated with a PR which was merged.
Shall we close the issue?
Or this merged code is still in dev?

akloeckner:
Same as in the other issue: we should soon merge the changes from dev to master. However, this issue here will not be closed automatically then, I think... I wonder, how we could manually put it on list of `semantic-release` for that...

ildar170975:
Shall we add a special label for it?

akloeckner:
Let's try, if that does something. :-D

ildar170975:
It will does. These labels may be used to track issues to be closed.

ildar170975:
@akloeckner 
Was it released?
See no "released" label here...

akloeckner:
I think, I got the "fixes some issue" syntax wrong and didn't properly connect this issue to the PR #882. :shrug:

I haven't seen this binning problem again. So I'll close the issue. 👍

ildar170975:
@akloeckner 
Also, there is an issue https://github.com/kalkih/mini-graph-card/issues/181 with a similar glitch: a left border is "bigger" than the right one.

PR #882, opened 2022-12-03
fix: add first datapoint tooltip for line graph

### LINKED ISSUE #879: First point not rendered as mouse-over-able
When we render a graph, the `line` is drawn for all coordinates (or, data points). But the first coordinate is not drawn as a `point`, i.e. the user cannot hover over the point and read its value.

Example following. Notice the missing small circle for the left-most data point. There should be 3 points. But only 2 are rendered.
image

Corresponding configuration:
```yaml
type: custom:mini-graph-card
entities:
  - entity: sensor.haushalt_gasverbrauch
hours_to_show: 3
points_per_hour: 1
aggregate_func: mean
hour24: true
decimals: 2
smoothing: false
show:
  graph: line
  labels: false
  state: false
```

Judging from the following code line, it seems this first point is removed on purpose. Is that correct? If so: Why is that? If not: Should it be fixed?
https://github.com/kalkih/mini-graph-card/blob/8e70381e3d887e5803bc8fc4bc115ecc4dc587e0/src/graph.js#L123

akloeckner:
Already the first commit 7e15867ed657cbd0a3b64a3a0a5d47c7f69ffb91 introducing tooltips did not render the first data point as tooltip-able. While there certainly was a reason for it, I cannot figure it out. Additionally, there is no such logic for bar tooltips (`getBars`) nor in the actual line graph (`getLine`). So, I propose to include the first point in the tooltips.

Fixes #879, #878

### DIFF
--- a/src/graph.js
+++ b/src/graph.js
@@ -118,16 +118,18 @@ export default class Graph {
       coords[1] = [this.width + this.margin[X], 0, coords[0][V]];
     }
     coords = this._calcY(this.coords);
-    let next; let Z;
-    let last = coords[0];
-    const coords2 = coords.map((point, i) => {
-      next = point;
-      Z = this._smoothing ? this._midPoint(last[X], last[Y], next[X], next[Y]) : next;
-      const sum = this._smoothing ? (next[V] + last[V]) / 2 : next[V];
-      last = next;
-      return [Z[X], Z[Y], sum, i];
-    });
-    return coords2;
+    if (this._smoothing) {
+      let last = coords[0];
+      coords.shift();
+      return coords.map((point, i) => {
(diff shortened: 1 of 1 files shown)


akloeckner:
Beware of #878. When trying to fix that, too, now, I realized they're linked somehow. Tooltip dates are off a day, now. I'll investigate and amend...

akloeckner:
Done. This might have fixed #878, too. Will need to check later...

akloeckner:
The smoothed graphs required some of the logic I had removed in my first commit. So, now cleaned up the change, such that the (somewhat involved) logic for smoothing is in a separate code block. This still fixes also #878.

Issue #944, opened 2023-03-25
Graph may not be updated


ildar170975:
HA 2023.3.5
latest `mini-graph-card`
Win10+Chrome

Consider this code:
```
input_number:
  testing_number_01:
    min: 0
    max: 1
    step: 1
    mode: slider
```
```
type: vertical-stack
cards:
  - type: entities
    entities:
      - entity: input_number.testing_number_01
        name: number
        secondary_info: last-changed

  - type: history-graph
    entities:
      - input_number.testing_number_01
    hours_to_show: 1

  - type: custom:mini-graph-card
    entities:
      - input_number.testing_number_01
    hours_to_show: 1
    points_per_hour: 3600
    aggregate_func: last
    smoothing: false
    line_width: 1
    cache: false
    show:
      labels: true

```
where `cache: false` seems to be not affecting at all, `points_per_hour: 240` is defined to ensure that changes made frequently by a user (here interval is 15 sec) are displayed properly.

Clear browser cache.

The initial state:
![image](https://user-images.githubusercontent.com/71872483/227688764-f1b03363-30b1-4ee4-80da-d702058e2e57.png)

Now change value and see how the `mini-graph-card` is updated.
**After 1 minute the graph is not updated:**
![image](https://user-images.githubusercontent.com/71872483/227688950-33b5f63b-e925-4b0d-a57f-af0f32bc3ac2.png)

Change again:
![image](https://user-images.githubusercontent.com/71872483/227688980-1c631bb2-91fd-44f6-bc67-f41db9ae4392.png)
Immediately after changing the graph is redrawn - **but we see only "0 -> 1" change instead of "0 -> 1 -> 0".**

Go to Editor - the graph is displayed properly:
![image](https://user-images.githubusercontent.com/71872483/227689050-1907629f-424f-4863-b7dc-e8b959fcd34b.png)

Go back, refresh (F5) - the graph is displayed properly:
![image](https://user-images.githubusercontent.com/71872483/227689116-cceb0488-b1c1-47b5-b6be-01cdc34889fc.png)

Setting `points_per_hour: 3600`, playing with `cache: false` do not help.

And now the weirdest part:
1. Copy the whole card onto another view.
2. Change value - the graph is updated properly - but after several minutes I observe the same behaviour.

jlsjonas:
I'll try to look into it this weekend, thnx for the report!

akloeckner:
When searching the repo for `update`, I found #250 . Maybe related.

[...]

Issue #935, opened 2023-03-05
too many updates since 2023.2?


rorso:
I recently found that my HASS home screen is quite sluggish and takes a lot of system resources.

I found the "culprit" with the mini-graph-card that is displaying a graph of my current energy consumption with three entities.
The sensor values are updated every 5 seconds and so the card draws about 200KB sensor data every 5 seconds by default because of "on every state change".

This may be because the new "streaming history" function implemented in the 2023.2 update, as I did not have this problem before that date.

While we CAN (and should) set the `update_interval` parameter in this case, it should not be necessary to poll new data more often than there are data points to plot.

So if I configure 12 `points_per_hour`, it seems to be sufficient to poll the history every 300 seconds after the first request to keep the graph clean, even if the data does change every 5 seconds - which is still displayed as the actual numerical value.

I propose an enhancement to not poll more often than `points_per_hour` as a default, in case `update_interval` is not set and not more often than the sensor value changes, in case it is quite the reverse.

MEKadan:
To add to this, the default for aggrecate_func is avg. I tried aggrecate_func: last, and sluggishness got much worse, to unbearable level. I also encountered browser crashes. Had to remove aggrecate_func.

I have a powerful PC, and still this. However, there are differences between browsers. FF and Vivaldi pretty much the same (bad). DuckDuckGo mobile a bit better. Brave seems to be OK. Did not test others

PR #961, opened 2023-04-26
fix: retain the last out-of-bounds state

### LINKED ISSUE #960: Cannot read properties of undefined (reading 'length')
0.12.0-dev.2

Observe this error in Browser console for ALL entities which were not changed within a selected period:

1) DB size did not change within 4 days (> 24 hrs which is a default value for `hours_to_show`)
```
type: custom:mini-graph-card
entities:
  - sensor.home_assistant_v2_db_size
```
изображение

изображение

2) This `input_number` helper was not changed for a long time.
Consider this card:
```
type: vertical-stack
cards:
    entities:
      - input_number.testing_device_class_apparent_power
  - type: history-graph
    entities:
      - entity: input_number.testing_device_class_apparent_power
  - type: entities
    entities:
      - entity: input_number.testing_device_class_apparent_power
```
Now the graph is not displayed, same errors in the console:
изображение
изображение

Change the helper's value - then the graph will be displayed w/o errors:

изображение

3. Any sensor which value was changed more than 15 minutes ago:
изображение

akloeckner:
This re-insures the logic to keep the last known state before the start of our grah interval within the history. While this is not really correct, it allows for rendering steady-state-graphs.

This element was mistakenly removed in #881.

The piece of code relies on history being chronographical.

fixes #960

### DIFF
--- a/src/graph.js
+++ b/src/graph.js
@@ -70,6 +70,8 @@ export default class Graph {
       const key = Math.floor(Math.abs(interval));
       if (!res[key]) res[key] = [];
       res[key].push(item);
+    } else {
+      res[0] = [item];
     }
     return res;
   }


akloeckner:
@ildar170975, can you run this fix on your end, too, to see if it really fixes the issue?

ildar170975:
@akloeckner 
Unfortunately, right now do not have a devcontainer to test this fix; could you create a js file - so I will replace it manually & test?

akloeckner:
I added the bundle js to the PR. I'll just have to remember to remove it again before merging...

ildar170975:
Thanks a lot !!!
Tested with new "mini-graph-card-bundle.js".
https://github.com/kalkih/mini-graph-card/issues/960 - seems to be fixed:
изображение

Surely more tests are required, will do it these days.

Also, (another issue probably) console log is flooded with "fetch finished" - is it expected?

akloeckner:
Thanks for testing! Let me know, when you're ok to merge! I'll create a new dev release then.

As to the fetches, that might be related to #935 ? I didn't notice that many requests yet. I think, there should be one request per graph.

ildar170975:
> Let me know, when you're ok to merge!

This particular issue (https://github.com/kalkih/mini-graph-card/issues/960) seems to be solved.
Currently I am mostly satisfied with the current implementation.
The only critical issue for me is https://github.com/kalkih/mini-graph-card/issues/944 - but again, not life-critical )))

akloeckner:
Ok, I'll merge this then. And will have a look at #944 later...

akloeckner:
I just noticed: This fix will not work, if the last change is even outside of the scope of the recorder component in home assistant. The HA history API just reports nothing in this case. I guess, there is not much we can do here. We can rather file an issue with HA to include such state in the history response, maybe.

akloeckner:
> I expect to see a flatline with a value "100" - for ANY `hours_to_show` (even > 240).

Exactly. But I seem to have no line at all in that case. In the dev tools of the browser, I have checked the response of the API request and it is simply `[]`. I have no easy and reliable way to reproduce, though.

ildar170975:
My test with some `input_number` which was changed a long time ago:

изображение

A bit "enriched" case:
изображение

akloeckner:
Hmm... Looks good to me. My data has also changed in the meantime, so... I guess, I'll observe and open a separate issue, once it happens again.

PR #997, opened 2023-08-10
chore: merge master

akloeckner:
This adds the additional commits from `master` into `dev`. That way, contributors shouldn't need to pay attention to the branch they are building their changes on. It should suffice then to just target the `dev` branch.


akloeckner:
As these commits are already out there, we'll have to merge them eventually. So, I'll go ahead with this `chore` PR.

PR #1041, opened 2023-12-05
fix wrong default value for value_factor in README.md

schmocker:
the default value for the parameter value_factor is documented as 0. But I think it should be 1.


ildar170975:
@schmocker 
Do not get disappointed & keep improving the `mini-graph-card` ))

schmocker:
@ildar170975 oh ok I see, you are correct. Well that it's even more confusing since this is not a 'factor' but actually an 'exponent' which together with base 10 builds the actual factor. I would recommend to let the users define the factors directly (not the exponent) or rename it into something like value_factor_exponent and explain it a bit better.

ildar170975:
> I would recommend to let the users define the factors directly (not the exponent) or rename it into something like value_factor_exponent and explain it a bit better.

You are right, could be confusing. Check https://github.com/kalkih/mini-graph-card/issues/591

Issue #894, opened 2022-12-22
documention: thresholds


nschloe:
The main documentation gives

```yaml
color_thresholds:
  - value: 0
    color: "#ff0000"
  - value: 1.333333
    color: "#ffff00"
  - value: 2.666667
    color: "#00ff00"
  - value: 4
    color: "#0000ff"
```

as an example for color thresholding. What exactly does this mean? 

- Between `0` and `1.3333` we have `#ff0000`? 
- What happens below `0`?
- What happens above `4`?

jlsjonas:
Hi, this should indeed be cleared up a little!

Essentially, the color_treshold value works like a minimum: if the graph value is higher than a threshold, its color will apply (highest gets precedence and the first color is used as the lowest color, regardless of its value)

> Between 0 and 1.3333 we have #ff0000?

Correct! 

> What happens below 0?

Currently, the first threshold color is used; although we might want to expose the default color if no threshold was reached in the future (to improve multi-graph & allow for "red alert only" thresholds)

> What happens above 4?

`#0000ff` is applied.


i.e.:
```yaml
color_thresholds:
  - value: 0 # currently ignored
    color: "#ff0000"
  - value: 1.333333 # anything above 1.33333 
    color: "#ffff00"
  - value: 2.666667
    color: "#00ff00"
  - value: 4
    color: "#0000ff"
```

nschloe:
If the the first "threshold" is ignored, then perhaps it should be there at all. That'd make it clearer that the thresholds are thresholds _from below_.

Lemonadel:
Hi, it looks like it’s broken or something for past few months? Or user is broken? :D 

My config (tried using code but indentation on mobile is garbage :/ )

no matter if I skip first value 0 thd it’s shown as orange all the time even when value < orange thd.

Code:
color_thresholds:
 - value: 0
    color: blue
 - value: 100
    color: blue
 - value: 200
    color: orange
 - value: 300
    color: red
hours_to_show: 1
hour24: True
points_per_hour: 60
show:
   legend: True
   icon: False

Result:

![IMG_2171](https://github.com/kalkih/mini-graph-card/assets/43881989/f08f98e2-6b79-4924-8e22-25ffea7569bc)

ildar170975:
First fix your posted code. Use triple "`".

Lemonadel:
Done, thank you.

ildar170975:
> color_thresholds

This is a global property, not per entity.
[...]

PR #1058, opened 2024-01-19
docs(color_thresholds): explain resulting colors

ildar170975:
https://github.com/kalkih/mini-graph-card/issues/894

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,10 +202,11 @@ color_thresholds:
   - value: 4
     color: "#0000ff"
 ```
-i.e. if value is between "0" (including this value) & "1.33333" - color is "#ff0000",
-between "1.33333" (including this value) & "2.666667" - color is "#ffff00",
-between "2.666667" (including this value) & "4" - color is "#00ff00",
-equal to or more than "4" - color is "#0000ff".
+The example above will result in the following colors of the graph: if value is 
+* between `0` (including this value) and  `1.33333`, the color is `#ff0000`,
+* between `1.33333` (including this value) and `2.666667`, the color is `#ffff00`,
+* between `2.666667` (including this value) and `4`, the color is `#00ff00`,
+* equal to or more than `4`, the color is `#0000ff`.
 
 As a shorthand, you can just use a color string for the stops that you want interpolated:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -202,6 +202,10 @@ color_thresholds:
   - value: 4
     color: "#0000ff"
 ```
+i.e. if value is between "0" (including this value) & "1.33333" - color is "#ff0000",
```
akloeckner, line 205: This looks quite "like a block", when rendered. Could you maybe make it a list?  I think, that would be more readable.

akloeckner (commented):
Thanks! Could you base this PR towards the `dev` branch, please? (Maybe it'll be time to merge `dev` into `master`, afterwards.)

I also added a comment to the new text.

README.md
ildar170975: Could you give an example?

README.md
```diff
@@ -202,6 +202,10 @@ color_thresholds:
...
+between "1.33333" (including this value) & "2.666667" - color is "#ffff00",
+between "2.666667" (including this value) & "4" - color is "#00ff00",
+equal to or more than "4" - color is "#0000ff".
```
akloeckner, line 208: I think, this change should make the resulting colors more readable:

```suggestion
The example above will result in the following colors of the graph: if value is 
* between `0` (including this value) and  `1.33333`, the color is `#ff0000`,
* between `1.33333` (including this value) and `2.666667`, the color is `#ffff00`,
* between `2.666667` (including this value) and `4`, the color is `#00ff00`,
* equal to or more than `4`, the color is `#0000ff`.
```

What do you think? (If I got it right, you can just "apply" or "accept" the suggestion, if you agree.)

akloeckner (commented):
I used the suggestion-feature to give an example for the list, I was refering to. I couldn't see this as rendered markdown in the comment, but I think, it should be the right syntax.

We  should also fix the PR title to something semantic, e.g. `docs(color_thresholds): explain resulting colors`. That way, we will have a semantic commit title, when we "squash and merge". And that will result in the release being bumped properly.

README.md
ildar170975: Good, I think

ildar170975: > fix the PR title 

Done!

akloeckner:
Looks great! I'll merge it. Thanks!

PR #1059, opened 2024-01-20
feat: pull two years worth of changes into release channel

akloeckner:
We've been fixing things in the `dev` branch for quite some while now. I think, it is time we merged the "recent" changes into our release channel, so everyone can use them.

@jlsjonas, what do you think? Anything to prepare from your point of view?


jlsjonas:
Hey, definitely good for me (2 years already? Damn ...  ).
If I remember correctly the actions should already handle preparation; worst case we'll have to update the version tag (as it's explicitly a -dev tag atm), will keep an eye on it.


No regressions compared to release, right?

ildar170975:
It's time I guess))

akloeckner:
Regressions, as in breaking changes? I'm not aware of any. At least not per the commit messages.

PR #1064, opened 2024-01-28
chore: archive build artefacts for non-dev testers

akloeckner:
@ildar170975 ;-)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,6 @@ jobs:
       - name: Archive distribution bundle
         uses: actions/upload-artifact@v4
         with:
-          name: dist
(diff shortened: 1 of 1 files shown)


akloeckner:
When this is merged, the test builds from the continuous integration become available from:
* Show all checks
* Click on "Details" of a "CI / build" job
* Expand step "Archive distribution bundle"
* Click the "Artifact download URL"

The artifacts include a zipped `mini-graph-card-bundle.js` which can be tested in home assistant.

akloeckner:
I notice, I missed chores in https://github.com/kalkih/mini-graph-card/blob/master/CONTRIBUTING.md

But as this is infrastructure only, I'll go ahead and merge. We can always revert this without users noticing.

PR #1065, opened 2024-01-28
fix: tooltip interval start could be after end

### LINKED ISSUE #181: Problems with "hours_to_show" = float value (particularly < 1.0)
I would like to be able to show 5 min cpu load graphs.
But this means the hours_to_show option should be of type real number (0.083333) instead of integer.
Could this be made possible perhaps?
Love this very nice card btw :)

akloeckner:
* basically: use milliseconds in constants rather than fractions of hours
* also remove strange intervall size of n_points if hours < 1 and points/hour < 1
* also use 1 minute offset for end only when group_by, because that is cosmetics actually

fixes #181

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -527,27 +527,34 @@ class MiniGraphCard extends LitElement {
 
   setTooltip(entity, index, value, label = null) {
     const {
+      group_by,
       points_per_hour,
       hours_to_show,
       format,
     } = this.config;
-    const offset = hours_to_show < 1 && points_per_hour < 1
-      ? points_per_hour * hours_to_show
-      : 1 / points_per_hour;
 
-    const id = Math.abs(index + 1 - Math.ceil(hours_to_show * points_per_hour));
+    // time units in milliseconds in this function
+    const interval = getMilli(1 / points_per_hour);
+    const n_points = Math.ceil(hours_to_show * points_per_hour);
+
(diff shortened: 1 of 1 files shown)


akloeckner:
@ildar170975 , could you please check, if this solves your issue? According to our contributing guidelines, I'd like to have your approval also for the cleanup part. So, please also check, if your other graphs (especially using group_by) are still fine.

ildar170975:
@akloeckner 
Is it this one?

image

ildar170975:
@akloeckner 
Tested with that js.
[Code from here.](https://github.com/kalkih/mini-graph-card/issues/181#issuecomment-1913344134)
Seems that the issue is fixed.

ildar170975:
But I see this glitch now - no date is shown:

image

akloeckner:
Yes, that's the download I meant. 👍 

Good to see, the original glitch seems fixed. I still don't fully understand why. I think it's something about multiplying small numbers (1/120h interval between points) with big ones (3600000ms in an hour). 🤷 

Let me know, when (and if) you're confident, this could be merged!

ildar170975:
> when (and if) you're confident, this could be merged!

Let's merge it... The more people will test it - the better.
In may old practice (cruise missiles from 1997) even a probably tiny fix needs a full-scope testing, testers should consider a product as a "black box". But this is not a "rocket science" ))

akloeckner:
:-) Agreed!

PR #1071, opened 2024-01-29
chore: direct dependabot to dev branch and group minor updates

akloeckner:
@jlsjonas, does that make sense?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,5 +8,11 @@ updates:
     target-branch: "develop"
   - package-ecosystem: "github-actions"
     directory: "/"
+    target-branch: "dev"
     schedule:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -8,5 +8,11 @@ updates:
     target-branch: "develop"
```
jlsjonas, line 8: Should match for both

jlsjonas (commented):
Definitely makes sense, thanks for looking into it!

akloeckner:
Should we also activate npm updates? Not sure. This will be quite a lot, I guess. Maybe grouping them might be good then, too?

Actually I think, it will be easier to keep them deactivated.

PR #1075, opened 2024-02-09
fix: always get points so state last works with other graphs

### LINKED ISSUE #736: Graph not rendering when using 'last' as show state
I have always increasing sensors for internet upload and download and I'm trying to make an hourly upload and download bar graph.  Everything works great except I'm trying to have the latest graph values show for the state rather than the actual sensor state values (as they are meaningless ... total upload and download since last router reboot).  If I change the show > state to true or false, everything is fine; it doesn't seem to like 'last' although that is the setting according to the docs which will render what I'm looking for.

I'm on the latest version of HA (2022.2.8), HACS, and the mini graph card.

```
type: custom:mini-graph-card
entities:
  - entity: sensor.asuswrt_download
    name: Download
    color: '#595BBD'
    show_state: true
  - entity: sensor.asuswrt_upload
    name: Upload
    color: '#CD90EA'
    show_state: true
show:
  state: last
  graph: bar
  labels: true
  legend: false
hours_to_show: 24
group_by: hour
aggregate_func: diff
name: Hourly Internet Bandwidth Used
icon: mdi:wan
```

This image is with the show > state set to true and everything looks good except the sensor states are showing.

image

jasonlewis:
Fixes #736 by always getting points.

There's probably a few ways to fix this as it seems the `state: last` functionality specifically relies on points being available, but points are only set when specifically showing points (and never for bar graphs).

I decided to always define points and adjust the logic for rendering points to be based on whether the config is set. Granted I haven't tested all scenarios for rendering points so this may not be an adequete solution in all cases.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -427,7 +427,7 @@ class MiniGraphCard extends LitElement {
   }
 
   renderSvgPoints(points, i) {
-    if (!points) return;
+    if (!this.config.show.points) return;
     const color = this.computeColor(this.entity[i].state, i);
     return svg`
       <g class='line--points'
@@ -759,6 +759,7 @@ class MiniGraphCard extends LitElement {
         if (!entity || this.Graph[i].coords.length === 0) return;
         const bound = config.entities[i].y_axis === 'secondary' ? this.boundSecondary : this.bound;
         [this.Graph[i].min, this.Graph[i].max] = [bound[0], bound[1]];
+        this.points[i] = this.Graph[i].getPoints();
         if (config.show.graph === 'bar') {
           const numVisible = this.visibleEntities.length;
           this.bar[i] = this.Graph[i].getBars(graphPos, numVisible, config.bar_spacing);
@@ -768,9 +769,6 @@ class MiniGraphCard extends LitElement {
           if (config.entities[i].show_line !== false) this.line[i] = line;
           if (config.show.fill
             && config.entities[i].show_fill !== false) this.fill[i] = this.Graph[i].getFill(line);
-          if (config.show.points && (config.entities[i].show_points !== false)) {
-            this.points[i] = this.Graph[i].getPoints();
-          }
           if (config.color_thresholds.length > 0 && !config.entities[i].color)
             this.gradient[i] = this.Graph[i].computeGradient(
               config.color_thresholds, this.config.logarithmic,


src/main.js
```diff
@@ -427,7 +427,7 @@ class MiniGraphCard extends LitElement {
   }
 
   renderSvgPoints(points, i) {
-    if (!points) return;
```
akloeckner, line 430: I don't think, this PR is complete. If we remove this statement, the passed parameter is useless and should also be removed. But I reckon the `points` passed into here are used elsewhere, too...

So, we should either `calcPoints` everywhere they are used. Or pass them around and make sure they are defined in the central place.

I'll have to look at this more closely, when I find the time.

akloeckner:
@jasonlewis, thanks for contributing! It seems we'll have to work on it a bit still. See my comment. Unfortunately, I will only be able to look into it in a few days...

src/main.js
jasonlewis: `points` is still used when redering the actual points as it is mapped further down in this function.

akloeckner:
I did think about this quite some time. Sorry for the delay.

Although it's a small change, I don't like that we would treat `points` differently from the other properties, such as `bar`. So, I wondered, could we make this more consistent and came up with two ideas:

1. What do you think about always defining all those properties and moving the conditions to the `render*` functions? I'm a bit hesitant, because I still don't fully understand, how much calculating this would cause.

2. One other thing, I just stumbled against is: we could maybe just use `this.Graph[id].coords[-1][V]` instead of `this.points[id][-1][V]`. Or would we abuse the `Graph` then?

What do you think? Maybe you discarded these ideas already for some reason?

Liquidmasl:
> I did think about this quite some time. Sorry for the delay.
> 
> Although it's a small change, I don't like that we would treat `points` differently from the other properties, such as `bar`. So, I wondered, could we make this more consistent and came up with two ideas:
> 
> 1. What do you think about always defining all those properties and moving the conditions to the `render*` functions? I'm a bit hesitant, because I still don't fully understand, how much calculating this would cause.
> 2. One other thing, I just stumbled against is: we could maybe just use `this.Graph[id].coords[-1][V]` instead of `this.points[id][-1][V]`. Or would we abuse the `Graph` then?
> 
> What do you think? Maybe you discarded these ideas already for some reason?

Your concern is valid, its a bit smelly to use a graph construct as data retrieval, but dont let a little smell stand in the way of a decent fix. This bug is real and the proposed fix works, and is just slightly smelly. 
Your second approach is probably cleaner, get the raw data with `this.Graph[id].coords[last][V]`. 

But this bug could have been fixed 2 years ago. A follow up issue could have been opened to refactor this so it doesnt fall through if the time for refactors arises. 

So my suggestion; merge this, make a follow up issue for a refactor. Get it fixed today, its waiting 2 years already for a minor smell.

Liquidmasl:
yeah well nvm, just close this PR in the favor or this one https://github.com/kalkih/mini-graph-card/pull/1308 and get it merged.

ildar170975:
Closed in favour of https://github.com/kalkih/mini-graph-card/pull/1322
Sincere thanks for contribution!

PR #1076, opened 2024-02-09
add: state sum to show sum of available points

jasonlewis:
I wanted a way to show the default state of the graph as a sum of the available points rather than the current/last point.

This adds a new `sum` value for `state` which sums the available points. This does rely on the fix within #1075 for it to work as it expects `points` to always be defined.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -277,7 +277,7 @@ class MiniGraphCard extends LitElement {
   getEntityState(id) {
     const entityConfig = this.config.entities[id];
     if (this.config.show.state === 'sum') {
-      return this.points[id].reduce((sum, point) => sum + point[V], 0)
+      return this.points[id].reduce((sum, point) => sum + point[V], 0);
     } else if (this.config.show.state === 'last') {
       return this.points[id][this.points[id].length - 1][V];
     } else if (entityConfig.attribute) {
--- a/src/main.js
+++ b/src/main.js
@@ -276,7 +276,9 @@ class MiniGraphCard extends LitElement {
 
   getEntityState(id) {
     const entityConfig = this.config.entities[id];
-    if (this.config.show.state === 'last') {
+    if (this.config.show.state === 'sum') {
+      return this.points[id].reduce((sum, point) => sum + point[V], 0)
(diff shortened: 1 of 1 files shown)


aaaaahaaaaa:
This is actually relevant to the issue is just opened: https://github.com/kalkih/mini-graph-card/issues/1095

I think your solution should be more holistics and cover all the `aggregate_func` types, not just `sum`. That feature would defo be a great addition.

przemek-baranowski:
My understanding is that the solution should work for all aggregation functions. It fetches values from previously generated points. "sum" is just another type of state; it does not correspond with any agg func. Prove me wrong or please merge this one. I'd be happy to use it ;)

ildar170975:
@jasonlewis 
Hi!
Can you tell me what is a use case for that feature?

Current situation:
1. Assume each point/bar is calculated based on "max/min/last/first/median/avg". Then a displayed state is either a current state, a last point/bar value (aggregated), or a selected point/bar value. In all cases values have similar physical meaning.
2. Assume each point/bar is calculated based on "sum/diff/delta".  Then a displayed state is either a current state, a last point/bar value (aggregated sum/delta/diff), or a selected point/bar value (aggregated sum/delta/diff). Here values have a different physical meaning - a current state is some "absolute" value, in other cases - that "sum/diff/delta" value.

Now you are proposing a new "SUM" value - which is a sum of all calculated points/bars.
This might have a sense in case of "max/min/last/first/median/avg", but not in case of " sum/delta/diff", right?
Assume that points/bars are calculated based "sum" aggregate_func. Then the "last" value will be your expected value, right?

jasonlewis:
@ildar170975 sorry it's been a while since this PR and I don't believe I'm using the card anymore or in the way I was wanting to. From memory though I was believe I was graphing kW produced by a solar system and each point was kW produced for an hour, and I was wanting a total of kW produced for that graph.

ildar170975:
@jasonlewis 
OK, no problem).
Consider closing this PR if you are not interested then.
May be someone else will come up with a similar or better idea...

Issue #1081, opened 2024-02-20
Show state next to a legend's label


ljmerza:
Is there a way to have the `show_state` next to the label instead of off to the side? With a few entities added, its not easily seem which state belongs to which entity

<img width="774" alt="image" src="https://github.com/kalkih/mini-graph-card/assets/4085765/de84ab9c-ce07-4d19-ae9c-70a1e3ccf61f">

ildar170975:
Similar proposal were here: https://github.com/kalkih/mini-graph-card/issues/679
Will keep both issues open:
1. this current issue is associated with a corresponding PR;
2. [another issue](https://github.com/kalkih/mini-graph-card/issues/679) contains card-mod-based solutions.

Probably both issues could be closed in case of the PR is reviewed & merged.

ildar170975:
Implemented in https://github.com/kalkih/mini-graph-card/pull/1173

PR #1082, opened 2024-02-20
feat: add show_legend_state

ljmerza:
https://github.com/kalkih/mini-graph-card/issues/1081

image

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -338,17 +338,29 @@ class MiniGraphCard extends LitElement {
 
   renderLegend() {
     if (this.visibleLegends.length <= 1 || !this.config.show.legend) return;
+
+
     return html`
       <div class="graph__legend">
-        ${this.visibleLegends.map(entity => html`
-          <div class="graph__legend__item"
-            @click=${e => this.handlePopup(e, this.entity[entity.index])}
-            @mouseenter=${() => this.setTooltip(entity.index, -1, this.getEntityState(entity.index), 'Current')}
-            @mouseleave=${() => (this.tooltip = {})}>
-            ${this.renderIndicator(this.entity[entity.index].state, entity.index)}
-            <span class="ellipsis">${this.computeName(entity.index)}</span>
-          </div>
-        `)}
+        ${this.visibleLegends.map((entity) => {
+    let legend = this.computeName(entity.index);
+    const state = this.getEntityState(entity.index);
+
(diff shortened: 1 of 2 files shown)


src/main.js
```diff
@@ -338,17 +338,29 @@ class MiniGraphCard extends LitElement {
...
-          <div class="graph__legend__item"
-            @click=${e => this.handlePopup(e, this.entity[entity.index])}
-            @mouseenter=${() => this.setTooltip(entity.index, -1, this.getEntityState(entity.index), 'Current')}
-            @mouseleave=${() => (this.tooltip = {})}>
-            ${this.renderIndicator(this.entity[entity.index].state, entity.index)}
-            <span class="ellipsis">${this.computeName(entity.index)}</span>
-          </div>
-        `)}
+        ${this.visibleLegends.map((entity) => {
+    let legend = this.computeName(entity.index);
```
akloeckner, line 346: Please properly indent the code.

src/main.js
```diff
@@ -338,17 +338,29 @@ class MiniGraphCard extends LitElement {
...
+    const state = this.getEntityState(entity.index);
+
+    const { show_legend_state } = this.config.entities[entity.index];
+    if (show_legend_state) {
+      legend += ` (${this.computeState(state)}${this.computeUom(entity.index)})`;
```
akloeckner, line 351: Can you put the state in an additional `span` with a suitable class? (If there already are suitable classes for state and UOM, you could maybe reuse them.)

I think there is a CSS property then to add the braces to the content. 

That way users could use `card-mod` to style the appearance of the state as they wish.

Please make sure the defaults look good in different situations, e.g. long and short state names.

akloeckner (changes requested):
Thanks for the suggestion! Looks like a useful feature.

I added a few remarks. I'll still have to test the feature before accepting it.

Please also direct your PR to the `dev` branch.

src/main.js
ildar170975: > put the state in an additional span

Absolutely correct.

Also, please think about an order: in some cases a user may want to have UoM before a value.

akloeckner: I think, a separate `span` should do the trick for that as well, together with the CSS `direction` property.

JulienDeveaux:
@akloeckner don't forget to close this one 😉

ildar170975:
Closed in favour of https://github.com/kalkih/mini-graph-card/pull/1173

PR #1105, opened 2024-05-15
fix: indicator style for the 1st state

ildar170975:
Consider the code:
```
type: custom:mini-graph-card
entities:
    show_state: true
    show_indicator: true
    show_state: true
    show_indicator: true
  - entity: sensor.mijia_300_1_co2
    show_state: true
    show_indicator: true
name: xxx
```
Currently an indicator for the 1st state is misaligned & square:

изображение

After the fix:

изображение

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,10 @@
+## [0.12.1](https://github.com/kalkih/mini-graph-card/compare/v0.12.0...v0.12.1) (2024-03-20)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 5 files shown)


ildar170975:
@akloeckner 
Have no idea why - but the PR contains some other changes as well:
изображение
although I only proposed changes in `style.js`.

akloeckner:
It's because I didn't pull the release-related commits into `dev`. I did that now. 

I'll fix the semantic PR title and merge. 👍

ildar170975:
Thanks a lot!

PR #1115, opened 2024-07-24
fix: legend indicators respect entity attribute configuration

arychj:
When using attributes for entity states, the line picks up the new state location and respects the color thresholds, however the legend indicators do not.

### Current behaivour
```yaml
entities:
  - entity: sensor.sensor_1
    attribute: attribute_1
    name: ' '
  - entity: sensor.sensor_2
    attribute: attribute_1
    name: ' '
# ...
```
image

### New behaviour
This updates the `renderIndicator()` function to use the existing `getEntityState()` which checks if an attribute has been configured, otherwise it falls back to the entity state.

image

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -345,7 +345,7 @@ class MiniGraphCard extends LitElement {
             @click=${e => this.handlePopup(e, this.entity[entity.index])}
             @mouseenter=${() => this.setTooltip(entity.index, -1, this.getEntityState(entity.index), 'Current')}
             @mouseleave=${() => (this.tooltip = {})}>
-            ${this.renderIndicator(this.entity[entity.index].state, entity.index)}
+            ${this.renderIndicator(this.getEntityState(entity.index), entity.index)}
             <span class="ellipsis">${this.computeName(entity.index)}</span>
           </div>
         `)}


ildar170975:
Hmmm, here it is said "attribute configuration".
I do not think I have recorded entities with numerical attributes, cannot test right now.
But there are issues with colored states, not attributes.
So let's start with a simpler case - state, not attribute.

1. No color thresholds:
```
type: custom:mini-graph-card
entities:
    show_state: true
    state_adaptive_color: true
    show_state: true
    state_adaptive_color: true
name: CO2
hours_to_show: 24
points_per_hour: 60
lower_bound: ~400
```
Here states & legend parts have correct colors:
изображение

2. Added `color_thresholds`:
```
color_thresholds:
  - value: 450
    color: orange
  - value: 550
    color: green
  - value: 650
    color: magenta
```
Here both entities are "pink" - states & legend parts - since both entities are > 650:
изображение

3. A bit different `color_thresholds` - 2nd entity is supposed to be below last threshold:
```
color_thresholds:
  - value: 450
    color: orange
  - value: 550
    color: green
  - value: 665    #!!!!!!!
    color: magenta
```
And here states have correct colors, but a legend part for the 2nd entity is black (WRONG):
изображение

akloeckner:
> So, imho in both cases - showing a state or an attribute - a legend parts should have colors dependently on their current value.

Thanks, that was what I was looking for. From your tests, it seems we still have some other issues. But this PR moves us in the right direction. 👍 

I'm going to merge it then.

ildar170975:
> I'm going to merge it then.

At least we'll be able to test it then )))

ildar170975:
@akloeckner
please tell me where to find a compiled js file

akloeckner:
I'll publish a pre-release in a few minutes. 

Also, I believe one of the actions creates a compiled js as an artifact. 🤔

akloeckner:
> Also, I believe one of the actions creates a compiled js as an artifact. 🤔

There should be one here: 
* Click on "Checks" for the PR
* Select "CI"
* Scroll down to "Artifacts"
* Download one of the zips
* It should contain the compiled .js

ildar170975:
After PR merged - became better, but a black indicator still shown: https://github.com/kalkih/mini-graph-card/issues/1116

PR #1118, opened 2024-07-26
fix: use d3 to interpolate colors

### LINKED ISSUE #1116: Legend indicator: color may not follow color_thresholds
Initial report: https://github.com/kalkih/mini-graph-card/pull/1115#issuecomment-2248682920

With https://github.com/kalkih/mini-graph-card/pull/1115 merged:
a legend indicator's color may follow `color_thresholds`:
```
type: custom:mini-graph-card
entities:
    show_state: true
    state_adaptive_color: true
    show_state: true
    state_adaptive_color: true
name: CO2
hours_to_show: 24
points_per_hour: 60
lower_bound: ~400
color_thresholds:
  - value: 450
    color: orange
  - value: 550
    color: green
  - value: 650
    color: magenta
height: 400
show:
  labels: true
```
изображение
i.e. here a color = PINK for both entities.

With changed `color_thresholds`:
```
color_thresholds:
  - value: 450
    color: orange
  - value: 550
    color: green
  - value: 680
    color: magenta
```
there is different picture - an indicator is BLACK for the 2nd entity:
изображение

Same - when graphs are built for attributes.

akloeckner:
This uses `d3-interpolate` to interpolate between two colors in the thresholds. The library can handle all sorts of color inputs, most importantly named colors. Our old implementation could only interpolate hexadecimal color values.

The following features now use the `d3` interpolation (and should thus be fixed):
* plot indicator relating a plot to the entity name
* filled plots with an area under the line

This also extends usage of color interpolation (to elements not interpolating at all before):
* the state
* points on the graph
* bars

Old (wrong though explicit) behaviour of bars can be restored by setting `color_thresholds_transition: hard`.

fixes #1116

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,12 +5,12 @@
   "requires": true,
   "packages": {
     "": {
-      "name": "mini-graph-card",
-      "version": "0.11.0",
+      "version": "0.12.2-dev.1",
       "license": "MIT",
       "dependencies": {
         "@kalkih/lz-string": "^1.4.5",
         "custom-card-helpers": "^1.6.6",
+        "d3-interpolate": "^3.0.1",
         "lit-element": "^2.2.1",
         "localforage": "^1.7.3",
         "spark-md5": "^3.0.1"
@@ -3905,6 +3905,25 @@
         "rollup": "^2.7.6",
         "superstruct": "^0.8.3",
         "typescript": "^3.8.3"
+      }
+    },
+    "node_modules/d3-color": {
+      "version": "3.1.0",
+      "resolved": "https://registry.npmjs.org/d3-color/-/d3-color-3.1.0.tgz",
+      "integrity": "sha512-zg/chbXyeBtMQ1LbD/WSoW2DpC3I0mpmPdW+ynRTj/x2DAWYrIY7qeZIHidozwV24m4iavr15lNwIwLxRmOxhA==",
+      "engines": {
+        "node": ">=12"
+      }
+    },
+    "node_modules/d3-interpolate": {
+      "version": "3.0.1",
+      "resolved": "https://registry.npmjs.org/d3-interpolate/-/d3-interpolate-3.0.1.tgz",
+      "integrity": "sha512-3bYs1rOD33uo8aqJfKP3JWPAibgw8Zm2+L9vBKEHJ2Rg+viTR7o5Mmv5mZcieN+FRYaAOWX5SJATX6k1PWz72g==",
+      "dependencies": {
+        "d3-color": "1 - 3"
+      },
+      "engines": {
+        "node": ">=12"
       }
     },
     "node_modules/dateformat": {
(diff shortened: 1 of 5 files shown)


akloeckner:
Example:
```yaml
    entities:
      - entity: sensor.aussentemperatur
        name: Außentemperatur
        unit: °C
      - entity: sensor.aussentemperatur
        name: Außentemperatur
        unit: °C
    hours_to_show: 72
    points_per_hour: 1
    group_by: hour
    aggregate_func: avg
    hour24: true
    decimals: 1
    show:
      labels: true
    color_thresholds:
      - value: 13
        color: blue
      - value: 20
        color: green
      - value: 23
        color: red
    color_thresholds_transition: smooth
```




| Before: | After: |
| --------| ---------|
| image | image|
| Indicator is black, blue is not used in Graph. | Correct indicator and graph. |

akloeckner:
I noticed, that bars do not inteprolate color. And there is some explicit code to prevent that. I wonder, if we should change that. See this example:

| Currently: | Could be: |
| ---- | ---- |
| image | image |

The original behavior could still be achieved by using `hard` color transition...

akloeckner:
I went for interpolating bar colors, too. It just doesn't make sense to not do it. This changes appearance of colored bars. But I consider it a fix. The old behaviour can be restored by setting `color_thresholds_transition: hard`.

akloeckner:
@ildar170975, could you please double-check that this behavior is as expected?

ildar170975:
@akloeckner 
Thanks a lot!
Just arrived to a normal PC, can test.
Please explain me once again (please....) where to find a js file to download...

akloeckner:
Don't hurry. I just noticed it was next week. ;-)

* Click on "Checks" for the PR
* Select "CI"
* Scroll down to "Artifacts"
* Download one of the zips
* It should contain the compiled .js

ildar170975:
> * It should contain the compiled .js

This one?
image

akloeckner:
Yes, any of the two dist* files should be ok. They are just compiled with different Node.js versions.

ildar170975:
@akloeckner 

Test 1:
```
type: custom:mini-graph-card
entities:
    show_state: true
    state_adaptive_color: true
    show_state: true
    state_adaptive_color: true
name: CO2
hours_to_show: 24
points_per_hour: 60
lower_bound: ~400
color_thresholds:
  - value: 450
    color: orange
  - value: 550
    color: green
  - value: 600
    color: magenta
height: 400
show:
  labels: true
```
image
Colors of legend's indicators are same as colors of state labels - as expected.

With attributes - colors are also OK (do not bother about absence of graphs since these "master" sensors are excluded from Recorder):
image

Imho we can merge this & allow other people to test.
I am having an access to PC only 3-4 days, then leaving for 8-10 days, and again 3-4 days, and so on - on a routine basis ))).

akloeckner:
> Imho we can merge this & allow other people to test.

Perfect. Let's do this then! Thanks for checking.


> I am having an access to PC only 3-4 days, then leaving for 8-10 days, and again 3-4 days, and so on - on a routine basis ))).

This sounds like an exciting life. Or an exhausting one. :-)

Issue #1047, opened 2023-12-30
min_bound_range ignores lower_bound


Tarudro:
Hi,

First of all, I love the card, thanks for this project and your effort!

However, I have noticed one thing that I can't really figure out and I believe it's kind of irritating as it is. (at least for my used case)

I have created mini-graph-cards to monitor my power consumptions. However, some devices have small values when in standby mode.
I've specified
`lower_bound: 0`
because I want to have the X axes stuck at the very bottom of the card, since I'm not expecting any negative values. Therefore a close to 0 consumption should also look like that. However, when you look at my screenshot, the "Washing Machine" doesn't look like that due to the small value scaling.

![image](https://github.com/kalkih/mini-graph-card/assets/4256079/6e3319ba-2f9e-4d83-83d5-d7b3295036ac)

I checked the options and found the  `min_bound_range:` setting.

But, when I specify this, the range neglects any defined lower_bound and creates the range to positive and negative side.

```
- type: custom:mini-graph-card
  icon: mdi:dishwasher
  name: Dishwasher
  hours_to_show: 6
  show:
    extrema: true
    legend: false
    average: true
    fill: fade
  line_width: 3
  lower_bound: 0
  min_bound_range: 10
  points_per_hour: 15
  entities:
    - entity: sensor.dishwasher_power
    - entity: sensor.dishwasher_energy_daily
      show_state: true
      show_graph: false
- type: custom:mini-graph-card
  icon: mdi:washing-machine
  name: Washing Machine
  hours_to_show: 6
  show:
    extrema: true
    legend: false
    average: true
    fill: fade
  line_width: 3
  lower_bound: 0
  points_per_hour: 15
  entities:
    - entity: sensor.washing_machine_power
    - entity: sensor.washing_machine_energy_daily
      show_state: true
      show_graph: false
```

I would suggest to make the following code change (I got no permissions to push it directly):
main.js
```
        boundary = [
          boundary[0] - diff / 2,
          boundary[1] + diff / 2,
        ];`
```
And apply the offset range only to the positive side, to keep the lower_bound.
![image](https://github.com/kalkih/mini-graph-card/assets/4256079/4952a355-a517-47fa-86b1-3608daf2ef0a)

ildar170975:
May I ask you why did you post:
- 3 unrelated screenshots
- a code for a card with no labels shown

?

This
> However, when you look at my screenshot, the "Washing Machine" doesn't look like that due to the small value scaling.

is rather unclear.

[...]

PR #1136, opened 2024-09-09
fix: more intuitive min_bound_range behavior

FAB1150:
This is a quick and simple change to the behavior of the min_bound_range option, to make it more intuitive to use and more predictable to new users.

Old behavior: it ignores every other bound option and overrides it

New behavior: If lower bound is set, it increases the range "upwards". If upper bound is set, it increases the range "downwards". 
If no bound is set, or if both upper AND lower bounds are set, old behavior is used.

This was mentioned in #1047, not an essential groundbreaking feature but it was a simple enough fix that I gave it a try. Tested on my machine.

This is my second ever PR so go easy on me please :)

Did I use the correct semantic title?

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -819,10 +819,16 @@ class MiniGraphCard extends LitElement {
 
       // Doesn't matter if minBoundRange is NaN because this will be false if so
       if (diff > 0) {
-        boundary = [
-          boundary[0] - diff / 2,
-          boundary[1] + diff / 2,
-        ];
+        if (min !== undefined && max === undefined) {
+          boundary[1] += diff;
+        } else if (max !== undefined && min === undefined) {
+          boundary[0] -= diff;
+        } else {
+          boundary = [
+            boundary[0] - diff / 2,
+            boundary[1] + diff / 2,
+          ];
+        }
       }
     }
 
--- a/src/main.js
+++ b/src/main.js
@@ -820,7 +820,7 @@ class MiniGraphCard extends LitElement {
       // Doesn't matter if minBoundRange is NaN because this will be false if so
       if (diff > 0) {
         const weights = [
-          min !== undefined && min[0] !== '~' ? 0 : (max === undefined ? 0 : 1),
+          min !== undefined && min[0] !== '~' || max === undefined ? 0 : 1,
           max !== undefined && max[0] !== '~' ? 0 : (min === undefined ? 0 : 1),
         ];
         const sum = weights[0] + weights[1];
--- a/src/main.js
+++ b/src/main.js
@@ -820,8 +820,8 @@ class MiniGraphCard extends LitElement {
       // Doesn't matter if minBoundRange is NaN because this will be false if so
       if (diff > 0) {
         const weights = [
-          min !== undefined && min[0] !== '~' ? 0 : 1,
-          max !== undefined && max[0] !== '~' ? 0 : 1,
+          min !== undefined && min[0] !== '~' ? 0 : (max === undefined && min !== undefined ? 0 : 1),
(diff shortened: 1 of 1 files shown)


akloeckner:
Thanks a lot! I had a look at the code, and I think it looks good. 👍 

Did you think about how soft bounds should best be handled? They're kind of in between `undefined` and fixed. Maybe they should be handled as if they were `undefined`?

I'll wait for @ildar170975's opinion, too, before merging. 

> Did I use the correct semantic title?

I would actually say, you're proposing a `fix`. You could edit the title to change this. A `refactor` would be, if you were changing the way different parts of the code interact with each other.

ildar170975:
@akloeckner 
I will be available in a week (I hope).

FAB1150:
@akloeckner 
I had thought of soft bounds, and didn't think it would have made a difference as this code gets executed later.
I didn't consider the fact that in the scenario that a soft bound is exceeded, my code still treats it as a bound and has a weird behavior.

I'll test your code and edit this comment when I do!

akloeckner:
> is this the behavior we want it to have?

To be agreed. I'd say, soft bounds are bound (😄) to be broken. So, they can also be extended by the minimum range. Whereas fixed bounds should stay fixed as long as possible.

But you're right. Maybe one soft bound plus one free bound: the soft bound should stay and the free one should be extended? So, my code is not optimal either.

FAB1150:
@akloeckner 
>the soft bound should stay and the free one should be extended?
That's what I was thinking, but maybe that's feature creep - your code looks good honestly (it just needed a !== instead of the != to avoid lint errors). I haven't really tested it yet as I was away, but it's compiled ready on my pc :)

FAB1150:
I thought about it for a bit, and came up with this:
```
const weights = [
  min !== undefined && min[0] !== '~' ? 0 : (max === undefined && min !== undefined ? 0 : 1),
  max !== undefined && max[0] !== '~' ? 0 : (min === undefined && min !== undefined ? 0 : 1),
];
```

Now if it's defined but a soft bound, and there is no other bound defined, it treats it as a bound. If both are soft bounds then it doesn't care. Tested it on my machine and it works great :)

I'll add this to the PR, let's see if we want to treat the soft bounds like this.

akloeckner (approved):
Thanks for looking into it!

Yes, it might be a bit looking too closely. On the other hand, I think, if we touch it, let's think it to the end.

Let's also still wait for @ildar170975's feedback. But for me it's a GO, now.

FAB1150:
@akloeckner tested that too, great! now it's more elegant.

akloeckner:
@ildar170975, I took the occasion and browsed through our open PRs. I said above, I wanted to wait for your feedback. Are you ok with the changes? (If you're tight on time, a superficial cross-check would be enough for me.)

ildar170975:
@akloeckner 
Yes, time is precious these days, I will sent you a PM in community, sorry for inability to close tasks rapidly...

ildar170975:
This code was added before this PR & it looks strange:
```
getBoundaries(series, min, max, fallback, minRange) {
  let boundary = [
   this.getBoundary("min", series, min, fallback[0], minRange),
   this.getBoundary("max", series, max, fallback[1], minRange),
  ];
```
because the `getBoundary()` does not accept the 5th parameter.

ildar170975:
Checked the added algorithm. Looks OK.
Let's fix [this](https://github.com/kalkih/mini-graph-card/pull/1136#issuecomment-2563191275) & then merge.

akloeckner:
Thanks a lot @FAB1150!

FAB1150:
thanks! first merged PR! :D

PR #1169, opened 2024-11-24
docs: Update README.md

seppelp:
Description of some properties where labeled with y-axis but they effect the x-axis.


ildar170975:
Wrong. These options are related to Y-axis, not to X-axis.

akloeckner:
There are two y axes: primary and secondary.

Issue #139, opened 2019-08-22
Feature request: align_unit: left or right


Proxx:
this feature request is not that big of a deal but it would be nice to aling the unit on the correct side of the state.

for example when we talk about euro's we use the notation like this "€ 20.50"  and not "20.50 €" like it is rendered now.

ildar170975:
@Proxx 
Check v0.14 - for monetary values, a locale-dependent presentation is used.
Perhaps you will find the "euro" symbol placed properly.

In general:
the `align_unit` option might be still useful anyway, this is put into consideration.

PR #1173, opened 2024-12-04
feat: add show_legend_state

JulienDeveaux:
Iterating on #1082 as I can't push into his repo

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,7 +130,7 @@ properties of the Entity object detailed in the following table (as per `sensor.
 | unit | string |         | Set a custom unit of measurement, overrides `unit` set in base config.
 | aggregate_func | string |         | Override for aggregate function used to calculate point on the graph, `avg`, `median`, `min`, `max`, `first`, `last`, `sum`.
 | show_state | boolean |         | Display the current state.
-| show_legend_state | boolean |  false  | Display the current state with the legend.
+| show_legend_state | boolean |  false  | Display the current state as part of the legend.
 | show_indicator | boolean |         | Display a color indicator next to the state, (only when more than two states are visible).
 | show_graph | boolean |         | Set to false to completely hide the entity in the graph.
 | show_line | boolean |         | Set to false to hide the line.
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ properties of the Entity object detailed in the following table (as per `sensor.
 | unit | string |  | Set a custom unit of measurement, overrides `unit` set in base config.
 | aggregate_func | string |  | Override for aggregate function used to calculate point on the graph, `avg`, `median`, `min`, `max`, `first`, `last`, `sum`.
 | show_state | boolean |  | Display the current state.
+| show_legend_state | boolean |  | Display the current state with the legend.
 | show_indicator | boolean |  | Display a color indicator next to the state, (only when more than two states are visible).
 | show_graph | boolean |  | Set to false to completely hide the entity in the graph.
 | show_line | boolean |  | Set to false to hide the line.
--- a/README.md
+++ b/README.md
@@ -123,24 +123,24 @@ properties of the Entity object detailed in the following table (as per `sensor.
 
 | Name | Type | Default | Description |
 |------|:----:|:-------:|-------------|
-| entity ***(required)*** | string |  | Entity id of the sensor.
-| attribute | string | | Retrieves an attribute or [sub-attribute (attr1.attr2...)](#accessing-attributes-in-complex-structures) instead of the state
-| name | string |  | Set a custom display name, defaults to entity's friendly_name.
-| color | string |  | Set a custom color, overrides all other color options including thresholds.
-| unit | string |  | Set a custom unit of measurement, overrides `unit` set in base config.
-| aggregate_func | string |  | Override for aggregate function used to calculate point on the graph, `avg`, `median`, `min`, `max`, `first`, `last`, `sum`.
-| show_state | boolean |  | Display the current state.
-| show_legend_state | boolean |  | Display the current state with the legend.
-| show_indicator | boolean |  | Display a color indicator next to the state, (only when more than two states are visible).
-| show_graph | boolean |  | Set to false to completely hide the entity in the graph.
-| show_line | boolean |  | Set to false to hide the line.
-| show_fill | boolean |  | Set to false to hide the fill.
-| show_points | boolean |  | Set to false to hide the points.
-| show_legend | boolean |  | Set to false to turn hide from the legend.
-| state_adaptive_color | boolean |  | Make the color of the state adapt to the entity color.
-| y_axis | string |  | If 'secondary', displays using the secondary y-axis on the right.
-| fixed_value | boolean |  | Set to true to graph the entity's current state as a fixed value instead of graphing its state history.
-| smoothing | boolean |  | Override for a flag indicating whether to make graph line smooth.
+| entity ***(required)*** | string |         | Entity id of the sensor.
+| attribute | string |         | Retrieves an attribute or [sub-attribute (attr1.attr2...)](#accessing-attributes-in-complex-structures) instead of the state
+| name | string |         | Set a custom display name, defaults to entity's friendly_name.
+| color | string |         | Set a custom color, overrides all other color options including thresholds.
+| unit | string |         | Set a custom unit of measurement, overrides `unit` set in base config.
+| aggregate_func | string |         | Override for aggregate function used to calculate point on the graph, `avg`, `median`, `min`, `max`, `first`, `last`, `sum`.
+| show_state | boolean |         | Display the current state.
+| show_legend_state | boolean |  false  | Display the current state with the legend.
+| show_indicator | boolean |         | Display a color indicator next to the state, (only when more than two states are visible).
+| show_graph | boolean |         | Set to false to completely hide the entity in the graph.
+| show_line | boolean |         | Set to false to hide the line.
+| show_fill | boolean |         | Set to false to hide the fill.
+| show_points | boolean |         | Set to false to hide the points.
+| show_legend | boolean |         | Set to false to turn hide from the legend.
+| state_adaptive_color | boolean |         | Make the color of the state adapt to the entity color.
+| y_axis | string |         | If 'secondary', displays using the secondary y-axis on the right.
+| fixed_value | boolean |         | Set to true to graph the entity's current state as a fixed value instead of graphing its state history.
+| smoothing | boolean |         | Override for a flag indicating whether to make graph line smooth.
 
 ```yaml
 entities:
--- a/src/main.js
+++ b/src/main.js
@@ -339,9 +339,8 @@ class MiniGraphCard extends LitElement {
   generateLegend(index) {
     let legend = this.computeName(index);
     const state = this.getEntityState(index);
-    const show_legend_state = this.config.entities[index];
 
-    if (show_legend_state) {
+    if (this.config.entities[index].show_legend_state) {
       legend += ` (${this.computeState(state)} ${this.computeUom(index)})`;
     }
 
--- a/src/main.js
+++ b/src/main.js
@@ -352,11 +352,12 @@ class MiniGraphCard extends LitElement {
   renderLegend() {
     if (this.visibleLegends.length <= 1 || !this.config.show.legend) return;
 
+    /* eslint-disable indent */
     return html`
       <div class="graph__legend">
         ${this.visibleLegends.map((entity) => {
-    const legend = this.computeLegend(entity.index);
-    return html`
+          const legend = this.computeLegend(entity.index);
+          return html`
             <div class="graph__legend__item"
               @click=${e => this.handlePopup(e, this.entity[entity.index])}
               @mouseenter=${() => this.setTooltip(entity.index, -1, this.getEntityState(entity.index), 'Current')}
@@ -365,9 +366,10 @@ class MiniGraphCard extends LitElement {
               <span class="ellipsis">${legend}</span>
             </div>
           `;
-  })}
+        })}
       </div>
     `;
+    /* eslint-enable indent */
   }
 
   renderIndicator(state, index) {
--- a/src/main.js
+++ b/src/main.js
@@ -339,7 +339,7 @@ class MiniGraphCard extends LitElement {
   generateLegend(entity, index) {
     let legend = this.computeName(index);
     const state = this.getEntityState(index);
-    const { show_legend_state } = this.config.entities[index];
+    const show_legend_state = this.config.entities[index];
 
     if (show_legend_state) {
       legend += ` <span class="state__value">(${this.computeState(state)}</span><span class="state__uom">${this.computeUom(index)})})</span>`;
--- a/src/main.js
+++ b/src/main.js
@@ -336,25 +336,29 @@ class MiniGraphCard extends LitElement {
       </div>` : '';
   }
 
+  generateLegend(entity, index) {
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -123,23 +123,24 @@ properties of the Entity object detailed in the following table (as per `sensor.
...
-| fixed_value | boolean |  | Set to true to graph the entity's current state as a fixed value instead of graphing its state history.
-| smoothing | boolean |  | Override for a flag indicating whether to make graph line smooth.
+| entity ***(required)*** | string |         | Entity id of the sensor.
+| attribute | string |         | Retrieves an attribute or [sub-attribute (attr1.attr2...)](#accessing-attributes-in-complex-structures) instead of the state
+| name | string |         | Set a custom display name, defaults to entity's friendly_name.
+| color | string |         | Set a custom color, overrides all other color options including thresholds.
+| unit | string |         | Set a custom unit of measurement, overrides `unit` set in base config.
+| aggregate_func | string |         | Override for aggregate function used to calculate point on the graph, `avg`, `median`, `min`, `max`, `first`, `last`, `sum`.
+| show_state | boolean |         | Display the current state.
+| show_legend_state | boolean |  false  | Display the current state with the legend.
```
ildar170975, line 133: This is the only required change, why other lines were changed? I see only whitespaces were added but it makes a "compare" view too crowded.

JulienDeveaux: When adding the false tag, my ide thought that it would look cleaner to have all blank spaces aligned.

ildar170975: Well, not critical. Main thing that the presentation should not be changed. Can't test myself, was it changed?

JulienDeveaux: Visually nothing, it's only how the file look when not rendered that has been changed

src/main.js
```diff
@@ -336,19 +336,35 @@ class MiniGraphCard extends LitElement {
       </div>` : '';
   }
 
+  generateLegend(entity, index) {
+    let legend = this.computeName(index);
+    const state = this.getEntityState(index);
+    const { show_legend_state } = this.config.entities[index];
+
+    if (show_legend_state) {
+      legend += ` <span class="state__value">(${this.computeState(state)}</span><span class="state__uom">${this.computeUom(index)})})</span>`;
```
ildar170975, line 345: Some uoms should be placed before a value - like "euro".

src/main.js
```diff
@@ -336,19 +336,35 @@ class MiniGraphCard extends LitElement {
       </div>` : '';
   }
 
+  generateLegend(entity, index) {
+    let legend = this.computeName(index);
+    const state = this.getEntityState(index);
+    const { show_legend_state } = this.config.entities[index];
```
ildar170975, line 342: I am not getting this line. I believed that `show_legend_state` is an boolean option.

JulienDeveaux: It's a leftover from the previous PR
It should indeed be a bool I'll change that

src/main.js
JulienDeveaux: Why euro should be before the unit ? I don't see it

ildar170975: You probably meant "before the state".
There are no strict rules in HA about Euro, I think.
If you create a `monetary` device_class it will show UoM as for other values:
image
Although on Energy page it will show it differently:

image
in fact, this "uom state" order in Energy is same for all currencies)))

Also, there was a FR: https://github.com/kalkih/mini-graph-card/issues/139
So, in future we should let users to define an order (not set a fixed order for "Euro" of course).

README.md
ildar170975: Suggest to change
"Display the current state with the legend."
to
"Display the current state as a part of the legend."

src/main.js
ildar170975: Also, could you explain a necessity of an `entity` variable here?
```
generateLegend(entity, index)
```

src/main.js
JulienDeveaux: I see
Maybe we should let this part for the #139 then ?

src/main.js
```diff
@@ -336,19 +336,35 @@ class MiniGraphCard extends LitElement {
...
-          <div class="graph__legend__item"
-            @click=${e => this.handlePopup(e, this.entity[entity.index])}
-            @mouseenter=${() => this.setTooltip(entity.index, -1, this.getEntityState(entity.index), 'Current')}
-            @mouseleave=${() => (this.tooltip = {})}>
-            ${this.renderIndicator(this.getEntityState(entity.index), entity.index)}
-            <span class="ellipsis">${this.computeName(entity.index)}</span>
-          </div>
-        `)}
+        ${this.visibleLegends.map((entity) => {
+    const legend = this.generateLegend(entity, entity.index);
```
ildar170975, line 357: See a remark about the `entity` variable above.

src/main.js
ildar170975: Yes, it should be a separate PR imho. Here we will use a fixed order "state uom".

src/main.js
JulienDeveaux: It's fixed

src/main.js
JulienDeveaux: The remark ?

JulienDeveaux:
In the previous PR, it was advised to add state__value & state__uom classes to the render.
It messes the visual as they are way too big for this small, so I removed them

src/main.js
ildar170975: It was deleted after commits.
Remark is that I see no need of passing `entities` variable into `generateLegend()`

JulienDeveaux: Indeed fixed

src/main.js
```diff
@@ -336,19 +336,35 @@ class MiniGraphCard extends LitElement {
       </div>` : '';
   }
 
+  generateLegend(entity, index) {
+    let legend = this.computeName(index);
+    const state = this.getEntityState(index);
+    const show_legend_state = this.config.entities[index];
```
ildar170975, line 342: Still not getting this.
You gave a bool `show_legend_state` option inside "entities" object.
Why not using it?
Instead you use `this.config.entities[index];` - which is an object, not a bool.

src/main.js
```diff
@@ -336,19 +336,35 @@ class MiniGraphCard extends LitElement {
...
+
+    if (show_legend_state) {
+      legend += ` (${this.computeState(state)} ${this.computeUom(index)})`;
```
ildar170975, line 345: For "%" units some people may say like "do not want to see a whitespace before %" (https://github.com/home-assistant/frontend/pull/19397). Personally me do not care.

ildar170975:
> so I removed them

totally agree, missed that myself.

src/main.js
JulienDeveaux: I didn't inspect what was really going on in this object
It should be better now

src/main.js
JulienDeveaux: I don't get this data in this.entity[index].attributes (tryed with system cpu usage metric)
I don't know if I can get this to work

ildar170975: This `this.config.entities[index].show_legend_state` may be not defined (and thus treated as `false`) - so please consider this case.

JulienDeveaux: If null, it should be considered as false.
I've added the check just in case, but it shouldn't be necessary

ildar170975: > I don't get this data in this.entity[index].attributes (tryed with system cpu usage metric)

Could you elaborate?

JulienDeveaux: I get the entity attributes in this.entity[index].attributes (here firendly_name, state_class, unit_of_measurement). I don't get any blank_before_percent attribute or something like that to detect whether I put the blank space

ildar170975: > I don't get any blank_before_percent attribute or something like that to detect whether I put the blank space

In HA frontend a special function is used to either set a blank or not between a state & unit, see the mentioned PR.
But let's not bother about these things in this PR ))

ildar170975: Please let me know if you done with changes.

JulienDeveaux: yeah i'm done :)

ildar170975:
@akloeckner 
Please have a look at this PR.
Cleaned up the changes together a bit.
I am not having enough experience to approve all changes.

What is not covered (and MAY BE will be fixed in other PRs):
1. Custom order "state uom" / "uom state". Currently it is fixed as "state uom" for a legend entry.
Note that for a "state" label a value & uom are kept in separate DOM elements - so an order can be changed by card-mod [as it was explained here](https://community.home-assistant.io/t/lovelace-mini-graph-card/71385/2185).
But in this PR a legend is presented as
```
legend += ` (${this.computeState(state)} ${this.computeUom(index)})`;
```
without keeping parts in separate elements. May be they should be placed both in `<span>` instead to allow users to use card-mod here.

2. There are [rules](https://github.com/home-assistant/frontend/pull/19397) in HA frontend regarding a whitespace between a value & uom. A special function is used for it. No idea if it can be used in custom cards or it should be repeated. Personally me do not care about whitespaces))).

src/main.js
```diff
@@ -336,19 +336,36 @@ class MiniGraphCard extends LitElement {
       </div>` : '';
   }
 
+  generateLegend(index) {
+    let legend = this.computeName(index);
+    const state = this.getEntityState(index);
+
+    const show_legend_state = this.config.entities[index].show_legend_state ?? false;
```
akloeckner, line 343: Personally, I liked the destructuring syntax for retrieving the `show_legend_state` option. This syntax also seems to be usually used in the existing codebase, whenever some property of `this.config` is simply assigned to a shortcut. So, I suggest to either revert to this syntax or directly go: 
```
if(this.config.entities[index].show_legend_state)
```

There is also a way to [assign a default value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#default_value), if the property is not set: 
```
const { show_legend_state = false } = this.config.entities[index];
```


(Also, I am not sure, if we support `??` in the `babel` configuration yet. It is not used in the codebase yet, I think.)

src/main.js
```diff
@@ -336,19 +336,36 @@ class MiniGraphCard extends LitElement {
...
-            @click=${e => this.handlePopup(e, this.entity[entity.index])}
-            @mouseenter=${() => this.setTooltip(entity.index, -1, this.getEntityState(entity.index), 'Current')}
-            @mouseleave=${() => (this.tooltip = {})}>
-            ${this.renderIndicator(this.getEntityState(entity.index), entity.index)}
-            <span class="ellipsis">${this.computeName(entity.index)}</span>
-          </div>
-        `)}
+        ${this.visibleLegends.map((entity) => {
+    const legend = this.generateLegend(entity.index);
+    return html`
```
akloeckner, line 359: Please indent properly.

src/main.js
```diff
@@ -336,19 +336,36 @@ class MiniGraphCard extends LitElement {
...
+            <div class="graph__legend__item"
+              @click=${e => this.handlePopup(e, this.entity[entity.index])}
+              @mouseenter=${() => this.setTooltip(entity.index, -1, this.getEntityState(entity.index), 'Current')}
+              @mouseleave=${() => (this.tooltip = {})}>
+              ${this.renderIndicator(this.getEntityState(entity.index), entity.index)}
+              <span class="ellipsis">${legend}</span>
+            </div>
+          `;
+  })}
```
akloeckner, line 368: Please indent properly.

src/main.js
```diff
@@ -336,19 +336,36 @@ class MiniGraphCard extends LitElement {
       </div>` : '';
   }
 
+  generateLegend(index) {
```
akloeckner, line 339: There is not yet a function named `generateXYZ` in the code. My feeling is, we use `computeXYZ` a lot for generating text. Should we rename `generateLegend` to `computeLegend`?

akloeckner (commented):
Thanks for sorting most things out already! And sorry for re-opening some of the topics. :-)

I agree with your general assessment:
* Let's keep the UOM order for later. I also understand from your conversation that the usual classes generated a strange look. So, it is more than just adding known classes and spans and should thus be treated with more thought in another PR.
* Same for the whitespace. If the state and UOM parts become separate tags with flexible order, it will be up to the user anyways to add or remove whitespace.

I haven't actually downloaded, compiled and tested the changes yet. I'll approve the workflow run, so maybe you can check it out, @ildar170975? You'll have to wait for the code to successfully compile...

src/main.js
JulienDeveaux: I tried but eslint throw an error if it is not like so.
To fix this, I've also tried to  ignore html templates, but it made the rest of the project wrong by eslint standard

src/main.js
JulienDeveaux: I've rolled back to the deconstructor + default value

src/main.js
JulienDeveaux: Same comment as above

src/main.js
JulienDeveaux: Renamed to computeLegend

src/main.js
akloeckner: That made me curious and I went down that painful road myself, too... I couldn't come up with something other than ignoring the rule for this single template using: 

```js
/* eslint-disable indent */
...
/* eslint-enable indent */
```

I would be fine to cheat eslint this time. What do you think?

src/main.js
JulienDeveaux: fixed

src/main.js
JulienDeveaux: fixed

src/main.js
```diff
@@ -336,11 +336,11 @@ class MiniGraphCard extends LitElement {
       </div>` : '';
   }
 
-  generateLegend(index) {
+  computeLegend(index) {
     let legend = this.computeName(index);
     const state = this.getEntityState(index);
 
-    const show_legend_state = this.config.entities[index].show_legend_state ?? false;
+    const { show_legend_state = false } = this.config.entities[index];
```
ildar170975, line 343: We have already discussed this earlier:
the `show_legend_state` is an option inside `config.entities`.
What is a purpose of assigning a bool value to an OBJECT?
Why not assigning it to `this.config.entities[index].show_legend_state` instead?

JulienDeveaux: See @akloeckner review that I have unresolved.
Personally, both are fine for me.

akloeckner: It's just a shorthand for the other version. 

This is equivalent: 
```js
a = config.a;
{a} = config;
```

So, actually, it's a matter of style. It has been used before in the project. That's why I suggested to use it again. 

Actually, the default value isn't used so far. But I think it cannot harm.

ildar170975: I have a very little experience in JS, so asking silly questions.
I believed that "this.config.entities[index]" is a particular member of "entities" list - i.e. it is an object containing all options inside a particular entity (including that "show_legend_state" option). But we need to take a particular "show_legend_state" option - that is why I believed that we need to write
```
const { show_legend_state = false } = this.config.entities[index].show_legend_state;
```
If I am wrong - let's finalize this PR then:
- generate js file for testing,
- test it together,
- then create a Release if everything is OK.

akloeckner: You're right about the structure of the config object. The "destructuring assignment" is just a short way to extract properties of the object into single variables. 

See the documentation here:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

akloeckner (approved):
I am fine with this PR. There is a js file to use for testing as usual in the "checks" section.

ildar170975:
Let's merge it. Looks great.

akloeckner:
Thanks for bearing with us @JulienDeveaux!

Issue #837, opened 2022-09-14
[Feature Request] Grid support


Hubertilus:
Hello,

I would like to have grid support for the graphs. Vertical and horizontal lines can help to identify how long the sensor was inside a specific range, or when a new day starts.

ildar170975:
This issue is about adding lines.
[Another issue is about adding labels.](https://github.com/kalkih/mini-graph-card/issues/739)

idogrf:
This would really be helpful

Brandon1212b:
Any update?

PR #1179, opened 2024-12-12
feat: add grid lines

JulienDeveaux:
Implements #837 

image
This is a grid_line_type of day for an hours_to_show of 730

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,7 @@ We recommend looking at the [Example usage section](#example-usage) to understan
 | state_map | [state map object](#state-map-object) |  | v0.8.0 | List of entity states to convert (order matters as position becomes a value on the graph).
 | value_factor | number | 0 | v0.9.4 | Scale value by order of magnitude (e.g. convert Watts to kilo Watts), use negative value to scale down.
 | logarithmic | boolean | `false` | v0.10.0 | Use a Logarithmic scale for the graph
+| grid_line_type | string |  | v0.12.x | Show grids lines using `hour` / `day` / `week` and the value of hours_to_show
 
 
 #### Entities object
--- a/src/main.js
+++ b/src/main.js
@@ -507,11 +507,12 @@ class MiniGraphCard extends LitElement {
   }
 
   renderSvg() {
-    const { height } = this.config;
+    const { height, grid_line_type = false } = this.config;
     return svg`
       <svg width='100%' height=${height !== 0 ? '100%' : 0} viewBox='0 0 500 ${height}'
         @click=${e => e.stopPropagation()}>
         <g>
+          ${grid_line_type ? this.renderGridLines() : ''}
           <defs>
             ${this.renderSvgGradient(this.gradient)}
           </defs>
@@ -523,6 +524,40 @@ class MiniGraphCard extends LitElement {
         </g>
         ${this.points.map((points, i) => this.renderSvgPoints(points, i))}
       </svg>`;
+  }
+
+  renderGridLines() {
+    const {
+      height, hours_to_show, grid_line_type, line_width,
+    } = this.config;
+
+    const containerWidth = 500;
+    let numLines;
+    let xRatio;
+
+    switch (grid_line_type) {
+      case 'hour':
+      default:
+        numLines = hours_to_show;
+        xRatio = containerWidth / numLines;
+        break;
+      case 'day':
+        numLines = Math.round(hours_to_show / 24);
+        xRatio = (containerWidth / numLines);
+        break;
+      case 'week':
+        numLines = Math.round(hours_to_show / 168);
+        xRatio = containerWidth / numLines;
+        break;
+    }
+    const lines = [];
+
+    for (let i = 0.5; i < numLines; i += 1) {
+      const x = xRatio * i;
+      lines.push(svg`<line x1=${x} y1="0" x2=${x} y2=${height} stroke="lightgray" stroke-width=${line_width}/>`);
+    }
+
+    return lines;
   }
 
   setTooltip(entity, index, value, label = null) {
--- a/src/main.js
+++ b/src/main.js
@@ -528,7 +528,7 @@ class MiniGraphCard extends LitElement {
 
   renderGridLines() {
     const {
-      height, hours_to_show, grid_line_type, line_width,
+      height, hours_to_show, grid_line_type,
     } = this.config;
 
     const containerWidth = 500;
@@ -552,9 +552,13 @@ class MiniGraphCard extends LitElement {
     }
     const lines = [];
 
-    for (let i = 0.5; i < numLines; i += 1) {
-      const x = xRatio * i;
-      lines.push(svg`<line x1=${x} y1="0" x2=${x} y2=${height} stroke="lightgray" stroke-width=${line_width}/>`);
+    for (let i = 0; i < numLines; i += 1) {
+      const x = xRatio * (i + 0.5);
+      if (i % 2 === 0) {
+        lines.push(svg`<line x1=${x} y1="0" x2=${x} y2=${height} stroke="lightgray" stroke-width="0.5" opacity="0.5"/>`);
+      } else {
+        lines.push(svg`<line x1=${x} y1="0" x2=${x} y2=${height} stroke="lightgray" stroke-width="0.5" opacity="1"/>`);
+      }
     }
 
     return lines;


README.md
```diff
@@ -115,6 +115,7 @@ We recommend looking at the [Example usage section](#example-usage) to understan
 | state_map | [state map object](#state-map-object) |  | v0.8.0 | List of entity states to convert (order matters as position becomes a value on the graph).
 | value_factor | number | 0 | v0.9.4 | Scale value by order of magnitude (e.g. convert Watts to kilo Watts), use negative value to scale down.
 | logarithmic | boolean | `false` | v0.10.0 | Use a Logarithmic scale for the graph
+| grid_line_type | string |  | v0.12.x | Show grids lines using `hour` / `day` / `week` and the value of hours_to_show
```
ildar170975, line 118: Let's move it inside a `show` option?

src/main.js
```diff
@@ -525,6 +526,44 @@ class MiniGraphCard extends LitElement {
       </svg>`;
   }
 
+  renderGridLines() {
+    const {
+      height, hours_to_show, grid_line_type,
```
ildar170975, line 531: f we decide to move grid_line_type into show option - this line should be changed

README.md
ildar170975: Also, is it possible to add `5minute` here? Note that `hours_to_show` could be fractional and < 1.

src/main.js
```diff
@@ -525,6 +526,44 @@ class MiniGraphCard extends LitElement {
...
+    } = this.config;
+
+    const containerWidth = 500;
+    let numLines;
+    let xRatio;
+
+    switch (grid_line_type) {
+      case 'hour':
+      default:
+        numLines = hours_to_show;
```
ildar170975, line 541: Note that `hours_to_show` could be fractional.

src/main.js
```diff
@@ -525,6 +526,44 @@ class MiniGraphCard extends LitElement {
...
+  renderGridLines() {
+    const {
+      height, hours_to_show, grid_line_type,
+    } = this.config;
+
+    const containerWidth = 500;
+    let numLines;
+    let xRatio;
+
+    switch (grid_line_type) {
```
ildar170975, line 538: Pretty please add `5minute` here )))

src/main.js
```diff
@@ -525,6 +526,44 @@ class MiniGraphCard extends LitElement {
...
+        numLines = Math.round(hours_to_show / 168);
+        xRatio = containerWidth / numLines;
+        break;
+    }
+    const lines = [];
+
+    for (let i = 0; i < numLines; i += 1) {
+      const x = xRatio * (i + 0.5);
+      if (i % 2 === 0) {
+        lines.push(svg`<line x1=${x} y1="0" x2=${x} y2=${height} stroke="lightgray" stroke-width="0.5" opacity="0.5"/>`);
```
ildar170975, line 558: The `lightgray` color looks differently in light/dark themes.
Let's try using a standard `--divider-color` variable instead.
This variable currently is `rgba(225,225,225,0.12)` for a default dark theme and `rgba(0,0,0,0.12)` for a default light theme.
Using `rgb(from var(--divider-color) R G B /0.5)` will change an opacity to `0.5` from `0.12` for this CSS variable.
So, we may use
```
stroke=var(--divider-color)
```
for more frequent lines and use
```
stroke=rgb(from var(--divider-color) R G B /0.5)
```
for less frequent lines; and the `0.5` value should be revised.

Also, we can make it even more flexible:
```
stroke=var(--mini-graph-card-small-grid-color, var(--divider-color))
...
stroke=var(--mini-graph-card-large-grid-color, rgb(from var(--divider-color) R G B /0.5))
```
i.e. incorporate own CSS variables (which may be "enreached" in future releases).

src/main.js
```diff
@@ -507,11 +507,12 @@ class MiniGraphCard extends LitElement {
   }
 
   renderSvg() {
-    const { height } = this.config;
+    const { height, grid_line_type = false } = this.config;
     return svg`
       <svg width='100%' height=${height !== 0 ? '100%' : 0} viewBox='0 0 500 ${height}'
         @click=${e => e.stopPropagation()}>
         <g>
+          ${grid_line_type ? this.renderGridLines() : ''}
```
ildar170975, line 515: Probably next "layers" will "overlay" the grid & make it less visible. I am thinking about drawing the grid AFTER these layers. But let's leave it this way & test...

src/main.js
```diff
@@ -525,6 +526,44 @@ class MiniGraphCard extends LitElement {
...
+
+    const containerWidth = 500;
+    let numLines;
+    let xRatio;
+
+    switch (grid_line_type) {
+      case 'hour':
+      default:
+        numLines = hours_to_show;
+        xRatio = containerWidth / numLines;
```
ildar170975, line 542: Shouldn't this line be placed outside `switch`? It is same for any case.

src/main.js
```diff
@@ -525,6 +526,44 @@ class MiniGraphCard extends LitElement {
...
+        break;
+      case 'week':
+        numLines = Math.round(hours_to_show / 168);
+        xRatio = containerWidth / numLines;
+        break;
+    }
+    const lines = [];
+
+    for (let i = 0; i < numLines; i += 1) {
+      const x = xRatio * (i + 0.5);
```
ildar170975, line 556: Assume we have `hours_to_show=2` & `grid_line_type=hour`.
Then `numLines  = 2` & `xRatio = 250`.
Then we will have 2 lines - for `125` & `375`, i.e. for `hour=0.5` & `hour=1.5`.
And the 1st line will be "thin" (0 % 2 === 0), the 2nd line will be "thick" (1 % 2 = 1).
Not nice.
ALternatively we can make "thin" lines for "hour=0.5" & "hour=1.5", and "thick" line for "hour=1".
What do you think?
Also, can we turn a "frequency" - that "2" value - into an option like `grid_line_?????` to define which lines should be "thin"?

src/main.js
```diff
@@ -507,11 +507,12 @@ class MiniGraphCard extends LitElement {
   }
 
   renderSvg() {
-    const { height } = this.config;
+    const { height, grid_line_type = false } = this.config;
```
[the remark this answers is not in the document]
JulienDeveaux: Done

src/main.js
JulienDeveaux: done

src/main.js
JulienDeveaux: done

src/main.js
JulienDeveaux: done

src/main.js
JulienDeveaux: I tried to move it around, but I don't see any visual changes to that

src/main.js
JulienDeveaux: done

src/main.js
JulienDeveaux: done

src/main.js
ildar170975: > don't see any visual changes to that

OK, we'll test it together

src/main.js
```diff
@@ -525,6 +527,49 @@ class MiniGraphCard extends LitElement {
...
+    }
+    const xRatio = containerWidth / numLines;
+    const lines = [];
+
+    for (let i = 0; i < numLines; i += 1) {
+      const x = xRatio * (i + 0.5);
+      if (i % grid_line_frequency > 0) {
+        lines.push(svg`<line x1=${x} y1="0" x2=${x} y2=${height} stroke="var(--divider-color)" stroke-width="0.5" opacity="0.5"/>`);
+      } else {
+        lines.push(svg`<line x1=${x} y1="0" x2=${x} y2=${height} stroke="rgb(from var(--divider-color) R G B /0.5)" stroke-width="0.5" opacity="1"/>`);
```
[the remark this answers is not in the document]
JulienDeveaux: Indeed, it's redundant
fixed

README.md
```diff
@@ -167,7 +167,8 @@ All properties are optional.
 | labels_secondary | `hover` | `true` / `false` / `hover` | Display secondary Y-axis labels.
 | name_adaptive_color | `false` | `true` / `false` | Make the name color adapt with the primary entity color.
 | icon_adaptive_color | `false` | `true` / `false` | Make the icon color adapt with the primary entity color.
-| grid_line_type | string |  | v0.12.x | Show grids lines using `5minute` / `hour` / `day` / `week` and the value of hours_to_show
+| grid_line_type | `hour` | `5minute` / `hour` / `day` / `week` | Show grids lines using  and the value of hours_to_show.
+| grid_line_frequency | 2 |  | Define the frequecy of thin / thick lines you want.
```
ildar170975, line 171: 1. Typo in "frequency".
2. Not sure about the "frequency" wording. May be "ratio"? And describe it like "Grid lines: thin/thick lines amount ratio". What do you think?

README.md
```diff
@@ -167,7 +167,8 @@ All properties are optional.
 | labels_secondary | `hover` | `true` / `false` / `hover` | Display secondary Y-axis labels.
 | name_adaptive_color | `false` | `true` / `false` | Make the name color adapt with the primary entity color.
 | icon_adaptive_color | `false` | `true` / `false` | Make the icon color adapt with the primary entity color.
-| grid_line_type | string |  | v0.12.x | Show grids lines using `5minute` / `hour` / `day` / `week` and the value of hours_to_show
+| grid_line_type | `hour` | `5minute` / `hour` / `day` / `week` | Show grids lines using  and the value of hours_to_show.
```
ildar170975, line 170: 1. The description has "using", is it left from some prev version?
2. "grids" -> "grid"

README.md
```diff
@@ -168,6 +167,8 @@ All properties are optional.
 | labels_secondary | `hover` | `true` / `false` / `hover` | Display secondary Y-axis labels.
 | name_adaptive_color | `false` | `true` / `false` | Make the name color adapt with the primary entity color.
 | icon_adaptive_color | `false` | `true` / `false` | Make the icon color adapt with the primary entity color.
+| grid_line_type | `hour` | `5minute` / `hour` / `day` / `week` | Show grid lines in function of hours_to_show value.
```
ildar170975, line 170: Let's put as "Show grid lines dependently on `hours_to_show` value."

README.md
```diff
@@ -168,6 +167,8 @@ All properties are optional.
...
+| grid_line_frequency | 2 |  | Grid lines: thin / thick lines amount ratio.
```
ildar170975, line 171: Please rename this parameter to "grid_lines_ratio". Also, lets have `grid_lines_type` instead of `grid_line_type` type )))

src/main.js
```diff
@@ -507,11 +507,13 @@ class MiniGraphCard extends LitElement {
   }
 
   renderSvg() {
-    const { height } = this.config;
+    const { height, show } = this.config;
+    const grid_line_type = show.grid_line_type ? show.grid_line_type : false;
```
ildar170975, line 511: grid_lines_type

src/main.js
```diff
@@ -507,11 +507,13 @@ class MiniGraphCard extends LitElement {
...
     return svg`
       <svg width='100%' height=${height !== 0 ? '100%' : 0} viewBox='0 0 500 ${height}'
         @click=${e => e.stopPropagation()}>
         <g>
+          ${grid_line_type ? this.renderGridLines() : ''}
```
ildar170975, line 516: grid_lines_type

src/main.js
```diff
@@ -525,6 +527,49 @@ class MiniGraphCard extends LitElement {
       </svg>`;
   }
 
+  renderGridLines() {
+    const {
+      height, hours_to_show, show,
+    } = this.config;
+    const grid_line_type = show.grid_line_type ? show.grid_line_type : false;
```
ildar170975, line 534: grid_lines_type

src/main.js
```diff
@@ -525,6 +527,49 @@ class MiniGraphCard extends LitElement {
...
+    const grid_line_frequency = show.grid_line_frequency ? show.grid_line_frequency : 2;
```
ildar170975, line 535: grid_lines_ratio

src/main.js
```diff
@@ -525,6 +527,49 @@ class MiniGraphCard extends LitElement {
...
+      case 'week':
+        numLines = Math.round(rounded_hours_to_show / 168);
+        break;
+    }
+    const xRatio = containerWidth / numLines;
+    const lines = [];
+
+    for (let i = 0; i < numLines; i += 1) {
+      const x = xRatio * (i + 0.5);
+      if (i % grid_line_frequency > 0) {
```
ildar170975, line 563: grid_lines_ratio

src/main.js
ildar170975: Shall we provide an additional safety that this value is numerical & integer?

JulienDeveaux: That's not a bad idea. I've added a isInteger check.

src/main.js
JulienDeveaux: done

src/main.js
JulienDeveaux: done

src/main.js
JulienDeveaux: done

README.md
JulienDeveaux: done

README.md
JulienDeveaux: done

src/main.js
```diff
@@ -525,6 +527,49 @@ class MiniGraphCard extends LitElement {
...
+        lines.push(svg`<line x1=${x} y1="0" x2=${x} y2=${height} stroke="var(--divider-color)" stroke-width="0.5"/>`);
+      } else {
+        lines.push(svg`<line x1=${x} y1="0" x2=${x} y2=${height} stroke="rgb(from var(--divider-color) R G B /0.5)" stroke-width="0.5"/>`);
```
ildar170975, line 568: Let's make it more complex - but also more flexible:
```
stroke=var(--mini-graph-card-large-grid-color, rgb(from var(--divider-color) R G B /0.5))
```

src/main.js
```diff
@@ -525,6 +527,49 @@ class MiniGraphCard extends LitElement {
...
+        numLines = Math.round(rounded_hours_to_show / 168);
+        break;
+    }
+    const xRatio = containerWidth / numLines;
+    const lines = [];
+
+    for (let i = 0; i < numLines; i += 1) {
+      const x = xRatio * (i + 0.5);
+      if (i % grid_line_frequency > 0) {
+        lines.push(svg`<line x1=${x} y1="0" x2=${x} y2=${height} stroke="var(--divider-color)" stroke-width="0.5"/>`);
```
ildar170975, line 566:
```
stroke=var(--mini-graph-card-thin-grid-color, var(--divider-color))
```

README.md
```diff
@@ -168,6 +167,8 @@ All properties are optional.
 | labels_secondary | `hover` | `true` / `false` / `hover` | Display secondary Y-axis labels.
 | name_adaptive_color | `false` | `true` / `false` | Make the name color adapt with the primary entity color.
 | icon_adaptive_color | `false` | `true` / `false` | Make the icon color adapt with the primary entity color.
+| grid_lines_type | `hour` | `5minute` / `hour` / `day` / `week` | Show grid lines in function of hours_to_show value.
+| grid_lines_ratio | 2 |  | Show grid lines dependently on `hours_to_show` value.
```
ildar170975, line 171: No , here there is a confusion:
1 - Show grid lines dependently on `hours_to_show` value.
2 - "Grid lines: thin / thick lines amount ratio"

src/main.js
JulienDeveaux: I don't see the added value here

README.md
JulienDeveaux: 🤦
My bad

src/main.js
```diff
@@ -507,11 +507,13 @@ class MiniGraphCard extends LitElement {
...
 
   renderSvg() {
-    const { height } = this.config;
+    const { height, show } = this.config;
+    const grid_lines_type = show.grid_lines_type ? show.grid_lines_type : false;
     return svg`
       <svg width='100%' height=${height !== 0 ? '100%' : 0} viewBox='0 0 500 ${height}'
         @click=${e => e.stopPropagation()}>
         <g>
+          ${grid_lines_type ? this.renderGridLines() : ''}
```
ildar170975, line 516: Regarding placing "".
In HA Frontend it is customary to use a `nothing` value.
In some my PR to HA frontend I also tried to use "" - but the told me that I must use `nothing`.
Honestly have no idea about a difference.
Do you know anything about it?

JulienDeveaux: Huh I've never heard about that, and it's not defined in js.
Setting it to undefined work, tho.

src/main.js
ildar170975: If CSS theme vars are used - users may customize a look dependently on their needs or their accessibility (a11y).
Customizing vars is possible either by custom themes or card-mod.

src/main.js
ildar170975: OK then, let's leave ""

src/main.js
JulienDeveaux: Good to know

ildar170975:
@JulienDeveaux 
The most difficult part is left - the algorithm itself.
Need some more time for thinking.
Thanks a lot for your work!

src/main.js
ildar170975: We still have an issue here.
If hours_to_show = 2, we will get the left picture (thick line on 125, thin on 375), but we need to get the right picture (thick line on 250):
image

JulienDeveaux: yeah that's the 0.5 in this line const x = xRatio * (i + 0.5);
If I remove it the first line is kinda hidden in the border and I find it wierd, that's why there is an offset

ildar170975: I am trying now to adapt the algorithm to a fractional `hours_to_show`. For int values already made it, but for fractional more difficult story.

README.md
```diff
@@ -168,6 +167,8 @@ All properties are optional.
 | labels_secondary | `hover` | `true` / `false` / `hover` | Display secondary Y-axis labels.
 | name_adaptive_color | `false` | `true` / `false` | Make the name color adapt with the primary entity color.
 | icon_adaptive_color | `false` | `true` / `false` | Make the icon color adapt with the primary entity color.
+| grid_lines_type | `hour` | `5minute` / `hour` / `day` / `week` | Show grid lines dependently on `hours_to_show` value.
+| grid_lines_ratio | 2 |  | Grid lines: thin / thick lines amount ratio
```
ildar170975, line 171: Please add a remark:
Grid lines: thin / thick lines amount ratio (0 - no thin lines)

src/main.js
```diff
@@ -525,6 +527,51 @@ class MiniGraphCard extends LitElement {
...
+    }
+    const xRatio = containerWidth / numLines;
+    const lines = [];
+
+    for (let i = 0; i < numLines; i += 1) {
+      const x = xRatio * (i + 0.5);
+      if (i % grid_lines_ratio > 0) {
+        lines.push(svg`<line x1=${x} y1="0" x2=${x} y2=${height} stroke="var(--mini-graph-card-thin-grid-color, var(--divider-color))" stroke-width="0.5"/>`);
+      } else {
+        lines.push(svg`<line x1=${x} y1="0" x2=${x} y2=${height} stroke="var(--mini-graph-card-thick-grid-color, rgb(from var(--divider-color) R G B /0.5))" stroke-width="0.5"/>`);
```
ildar170975, line 568: Please replace CSS vars' names:
--mini-graph-card-thick-grid-color ->
--mcg-grid-line-thick-color

same for "thin" (just realized we already have `--mcg-title-letter-spacing` & `--mcg-title-font-size` supported)

ildar170975: Also, please add at the end of "Theme variables" chapter of Docs these lines
| mcg-grid-line-thick-color | rgb(from var(--divider-color) R G B /0.5) | Grid "thick" line color.
| mcg-grid-line-thin-color | var(--divider-color) | Grid "thin" line color.

README.md
JulienDeveaux: done

src/main.js
JulienDeveaux: Done.
I'm really not familiar with available css global variables.
Do you know a place where they are centralized ?

ildar170975: There are plenty of css vars:
- added by 3rd party components used in HA (like Material) - example `--mdc-text-field-label-ink-color`;
- own HA frontend vars - example `--primary-text-color`;
- added by custom cards like mini-graph-card - example `--mcg-title-letter-spacing`;
- and users may add personal vars.

There is not place where all vars are centralized.
If you are looking for a tutorial related to CSS modding of HA frontend - check this [main card-mod thread](https://community.home-assistant.io/t/card-mod-add-css-styles-to-any-lovelace-card/120744) -> 1st post -> link at the bottom titled "fantastic" -> you will get a set of tutorials & examples.

ildar170975:
@JulienDeveaux 
Please take this C-like code and refactor to JS:
```
    float numLines;
    float spanFactor;
    float spanInHours;
    
    switch (grid_lines_type) {
      case "5minute":
        spanInHours = float(1/12);
        break;
      case "hour":
        spanInHours = 1;
        break;
      case "day":
        spanInHours = 24;
        break;
      case "week":
        spanInHours = 168;
        break;
      default:
        spanInHours = 1;
    }

    numLines = hours_to_show / spanInHours;
    spanFactor = ceil(hours_to_show / spanInHours) / (hours_to_show / spanInHours);
    float thickPart = containerWidth * spanFactor / ceil(numLines);
    float thinPart = thickPart / float(grid_lines_ratio + 1);
    numLines = numLines * (grid_lines_ratio + 1);

    for (int i = 0; i < numLines - 1; i++) {
      float x = containerWidth - thinPart * (i + 1);
      //float timeLabel = hours_to_show / numLines * (i + 1);
	
      if ((i + 1) % (grid_lines_ratio + 1) > 0) {
        // draw thin
      } else {
        // draw thick
      }
    }
```
The code was tested locally.
Leave the "float timeLabel" line commented - we may need it in future if decide to draw labels on X-axis.

src/main.js
ildar170975: Please check my addition below (currently last comment in the thread).

src/main.js
JulienDeveaux: I'll take a look thanks

src/main.js
JulienDeveaux: refactored

src/main.js
```diff
@@ -525,6 +527,54 @@ class MiniGraphCard extends LitElement {
...
+    const thickPart = containerWidth * spanFactor / Math.ceil(numLines);
+    const thinPart = thickPart / (grid_lines_ratio + 1);
+    numLines *= (grid_lines_ratio + 1);
+
+    const lines = [];
+    for (let i = 0; i < numLines - 1; i += 1) {
+      const x = containerWidth - thinPart * (i + 1);
+      // const timeLabel = hours_to_show / numLines * (i + 1);
+
+      if ((i + 1) % (grid_lines_ratio + 1) > 0) {
```
ildar170975, line 568: Can we simplify this?
Untested (please check if it OK):
```
      let stroke;
      if ((i + 1) % (grid_lines_ratio + 1) > 0) {
        stroke = "var(--mcg-grid-line-thin-color, var(--divider-color))";
      } else {
        stroke = "var(--mcg-grid-line-thick-color, rgb(from var(--divider-color) R G B /0.5))";
      }
      lines.push(svg`<line x1=${x} y1="0" x2=${x} y2=${height} stroke="${stroke}" stroke-width="0.5"/>`);
```

src/main.js
ildar170975: Very good.
Look, I have a little experience in JS - so have lots of doubts.
In C we need to provide things like
```
foo = a / float(b);
```
to avoid undesired conversions.
Is it properly handled in JS? Don't we need to write additional checks or explicit conversions ?

JulienDeveaux: The type in js is "number", it can be either int, float, long ... 
Here, js automatically assign the digits (if you want an int for example, you need to explicitly remove the digits)

src/main.js
JulienDeveaux: Seems to work fine

src/main.js
ildar170975: So, no additional conversions are needed in this code? Will it not loose a precision? Critical for the algorithm.

ildar170975:
@akloeckner 
Please have a look at this PR.

Guys, I am away from PC, hope will be back in a week. Meanwhile please test this PR.

src/main.js
JulienDeveaux: I've double-checked
js is not the best when a LOT of digits are involved. Here it is not a big deal as it is only light rounding problems.
So there is no explicit conversions needed.

ildar170975: Ok. Let us proceed with testing with a generated js-file. Please participate if possible).

JulienDeveaux: I'm currently using my 3 PRs in my daily driving dashboard. Is that what you mean by participate in the testing ?

ildar170975: Do you mean PRs after all our joint corrections?

JulienDeveaux: yeah I did a merge on a local branch this weekend, built it and pushed it to my HA.
For this one and the labels, It's rocking.
For the loader, I might come back to the code to check if my condition to show to loader is right.

ildar170975: Good, at least grid lines and legend labels works fine. Hope will test them too in a few days. These are very important PRs.

README.md
```diff
@@ -168,6 +167,8 @@ All properties are optional.
 | labels_secondary | `hover` | `true` / `false` / `hover` | Display secondary Y-axis labels.
 | name_adaptive_color | `false` | `true` / `false` | Make the name color adapt with the primary entity color.
 | icon_adaptive_color | `false` | `true` / `false` | Make the icon color adapt with the primary entity color.
+| grid_lines_type | `hour` | `5minute` / `hour` / `day` / `week` | Show grid lines dependently on `hours_to_show` value.
```
ildar170975, line 170: Let's change it to:
grid_lines_type | `hour` | `false`/`5minute` / `hour` / `day` / `week` | Show grid lines dependently on `hours_to_show` value (set to `false` or omit this option if grid lines should not be displayed).

akloeckner: It should. However, it could be simplified. ;-)

Also, I think, we should consider to have an option of grid_lines_per_hour, rather than grid_lines_type. It would allow for more flexibility with much simpler code. But I haven't thought this to the end yet.

src/main.js
ildar170975, line 516: Let's place all grid lines into a separate `<g>` element with a defined class:
1. Move this line BEFORE the `<g>` line (like it is done for `renderSvgPoints`).

2. In `renderGridLines` please try doing this (taken from `renderSvgPoints`):
```
    return svg`
      <g class='grid-lines'
        ... show lines
      </g>`;
```
The goal is to place all `<line>` elements into a separate `<g>` element with a defined `grid-lines` class. This will allow users to easier use card-mod; also this will allow developers to debug the appearance of the card in Code Inspector.

If this is hard to implement - please tell.

README.md
ildar170975: > grid_lines_per_hour, rather than grid_lines_type

The initial idea was:
-- "main" (thick) lines are shown every 5 minutes / hourly / daily / ... ;
-- "secondary" (thin) lines are shown more frequently ("ratio" option) or not shown at all.
I honestly cannot see how `grid_lines_per_hour` may be used along with some "show secondary lines" option (or I am still far away from HA & normal life, need more time)

src/main.js
ildar170975: @JulienDeveaux 
@akloeckner 
Guys, let's discuss this.
Currently parts of a graph are shown in this order:
-- grid lines
-- graph curve (class `line--rect`) + fill (class `fill--rect`)
-- graph points (class `line--points`)
Check this - the grid lines are below a curve:
image
and it definitely below a fill which has some transparency - so we can see grid line under the fill.
So far it looks more or less great.

But if a user makes a fill opaque:
image
with
```
        card_mod:
          style: |
            .fill {
              opacity: 1 !important;
            }
```
then the grid lines are not visible.

What we can do - we can place the grid lines right before the `<g.line--points>` element (it must capture mouse events).
So, the grid lines will be above the curve, above the fill - but below the points.

Not sure that this look will be great.
IMHO the best look will be this:
-- fill
-- grid lines
-- curve
-- points
But in this case we have to place all these `<line>` elements inside the same `<g>` element (like it is already done currently) - i.e. not in the separate `<g.grid-lines>` element how I suggested.
Or (just a guess) we can try splitting the "main" `<g>` element into separate parts for the curve & the fill.

What do you think?

README.md
akloeckner: Just from the top of my head, I was thinking: 
* Gridlines per hour is basically the same as proposed now. We could simply use fractions to yield days or weeks. But we'd be able to also have a gridline each second if need be.
* (It would also be consistent with the default `points_per_hour` setting. And reduce repeated code to actually convert days or weeks into multiples of hours.)
* We could also omit the secondary grid lines as different "class". And use `nth-of-type` in css instead. Again it reduces code and is more customizable by the user.

src/main.js
JulienDeveaux: Done

src/main.js
akloeckner: Technically, I would consider grid lines to be part of the plot's background, i.e. they would be drawn below everything else. Even opaque fills should still benefit from such lines, because we can still see them, where they "stick out" on top of the fill.

I just checked out a bit, how professional plot libraries like MATLAB, Python or pgfplots handle this. (Hoping to find some consensus standard to just follow.) And Python's matplotlib is the only candidate to put grid lines in the foreground. The others seem to put them in the background. So, no consensus, unfortunately. 🤷

src/main.js
```diff
@@ -574,7 +574,7 @@ class MiniGraphCard extends LitElement {
       lines.push(svg`<line x1=${x} y1="0" x2=${x} y2=${height} stroke="${stroke}" stroke-width="0.5"/>`);
     }
 
-    return svg`<g>${lines}</g>`;
+    return svg`<g class="grid-lines">${lines}</g>`;
```
ildar170975, line 577: Have no idea why - but classes in svg are named with 3 dashes:
image
so let's blindly change ours to `grid--lines` ))))

JulienDeveaux: huh ok why not

src/main.js
ildar170975: > where they "stick out" on top of the fill.

I think this is a proper way.
Currently our grid lines are below the fill.
So, to achieve this we will have to split the "main" `<g>` into "fill" & "line" parts - and then place the new `grid--lines` element between them.
This is a content of `<g>` for 1 entity:
image
I honestly have no idea what these `defs` are for.

akloeckner: Just guessing: the Def's are the colour map. That might also be the reason why fill and line are in the same group (`<g>`).

ildar170975: Do you mean `color_thresholds`? One `defs` for a curve, another one for a fill.

akloeckner: Also, looking at the code, we could simply render grid lines into the main group in renderSvg. We could then choose the order easily. Or is there a need to place them before the group itself as proposed currently?

However, by "stick out" I meant: the grid is hidden behind the fill. But it sticks out between the header and the fill. So, we can still use grid lines for orientation, but they do not disturb the clean look of the graph too much.

ildar170975: > render grid lines into the main group in renderSvg

Grid lines were inside the main `<g>`. I asked to place them into a separate `<g>`.

ildar170975: > the grid is hidden behind the fill. But it sticks out between the header and the fill. So, we can still use grid lines for orientation, but they do not disturb the clean look of the graph too much.

I see, this could be a better way - from some point of view.
Let's keep it like it is already done:
-- grid lines in a separate `<g>`
-- this group is before the main `<g>`.
And we will may be improve it in future.

akloeckner: > Do you mean `color_thresholds`? One `defs` for a curve, another one for a fill.

Yes. That's what I thought. But it's not really relevant to this PR if we stick to the order we have it right now.

> Let's keep it like it is already done:

Alright! :+1:

ildar170975: @JulienDeveaux
Do you have something to add? Or we "resolve the conversation"?

JulienDeveaux: That's good for me

README.md
ildar170975: Please let me think a bit.

ildar170975: @akloeckner 
Was thinking about using `grid_lines_per_hour `:

1. Currently we got new proposed options - `grid_lines_type` (`false` / `5minute` / `hour` / `day` / `week`) & `grid_lines_ratio` (integer, >=0).
Alternatively, we could get smth like `grid_lines_per_hour` (for thick lines) and same `grid_lines_ratio` (for thin lines).
And in this case this `grid_lines_per_hour` is numerical (could be fractional < 1).
Let's consider the "<1" case - it is needed to show thick lines, for example, every 2 hours, every day etc.
For the "every day" case a user will have to use a `0.041667`value.
Does not look nice, a `1/24` value seems to be more user-friendly - but in this case we need to process these `1/x` values and convert them to valid floats.

2. If we choose the current implementation with `grid_lines_type` - we will have a minimal granularity for thick lines = 5 minutes.
A smaller granularity then could be achieved by `grid_lines_ratio`:
5 - thin lines every 1 minute,
10 - every 30 seconds, and so on.
Probably this would be sufficient for most scenarios.

> We could also omit the secondary grid lines as different "class". And use nth-of-type in css instead. Again it reduces code and is more customizable by the user.

Sorry, I think using `nth-of-type` will make thing more difficult.
Assume we got a set of thick & thin lines WITHOUT classes.
To style them, we could use `nth-of-type` or similar - but a formula could be complex.
We draw these lines from a right side (now -> past). So, we will have to process these `nth` from the end as well. Not easy I think.
Using a separate class for thick & thin lines will allow users to quickly show/hide them by card-mod.

JulienDeveaux:
@ildar170975 @akloeckner 
Do you guys think this is ready, or it's still missing something ?

ildar170975:
I think it is good. But @akloeckner shared some thoughts, waiting for his feedback. 

Guys, I am far from PC these days. Cannot test anything.

ildar170975:
@JulienDeveaux 
@akloeckner 
BTW, a useful remark: https://github.com/kalkih/mini-graph-card/pull/1210#issuecomment-2646673927

PR #1193, opened 2025-01-02
fix: remove unused argument from getBoundary

ildar170975:
https://github.com/kalkih/mini-graph-card/pull/1136#issuecomment-2563191275


akloeckner:
I'm curious how that will work. :-)

akloeckner (approved):

akloeckner:
Like a charm... :-)

FAB1150:
thanks! Doing a quick blame, it seems like this was introduced back in #408. nice that it was such a simple fix :)

Issue #1195, opened 2025-01-06
Cannot hide a graph for entity


Koky05:
After last update of HA to 2025.01 functionality for hide graph for some entity stop working. It show only load icon:
![Screenshot_2025-01-06-01-19-58-23_c3a231c25ed346e59462e84656a70e50.jpg](https://github.com/user-attachments/assets/50565537-2365-405d-934d-d83fc47a5cb8)

I use it with next code, to have summary label with detail graphs:

```
  - type: custom:mini-graph-card
    name: Teplota vonku
    entities:
      - entity: sensor.teplota_vonku_2
        show_state: true
        show_graph: false
      - entity: sensor.broadcast_outsidetemp_temp2
        name: Vaillant
      - entity: sensor.thm_139335746
        name: Jablotron vonku
      - entity: sensor.thm_137742250
        name: Búda Hugo
    points_per_hour: 6
    show:
      labels: false
```

ildar170975:
This is how the code was posted, please correct.
![image](https://github.com/user-attachments/assets/cdb8f172-4641-48c8-99a8-1720d58bead3)

ildar170975:
OK, here is a short code to demonstrate the bug:
```
type: vertical-stack
cards:
  - type: custom:mini-graph-card
    entities:
      - entity: sensor.xiaomi_cg_1_co2
        show_state: true
        show_graph: false
      - entity: sensor.xiaomi_cg_2_co2
  - type: custom:mini-graph-card
    entities:
      - entity: sensor.xiaomi_cg_1_co2
        show_state: true
      - entity: sensor.xiaomi_cg_2_co2
```
![image](https://github.com/user-attachments/assets/ebae81d9-d2aa-4823-af4f-e0fc82eb3637)

The bug was added by https://github.com/kalkih/mini-graph-card/pull/1180

ildar170975:
@JulienDeveaux
Can you check this line:
https://github.com/kalkih/mini-graph-card/blob/f5261d92c37b93f2fd8c7f8da01fe8af4a5a18e8/src/main.js#L326

PR #1197, opened 2025-01-11
fix: graph loading indicator

ildar170975:
1. Added `show::loading_indicator` option (for users who do not like it).
2. Fix for https://github.com/kalkih/mini-graph-card/issues/1195.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -323,15 +323,15 @@ class MiniGraphCard extends LitElement {
   }
 
   renderGraph() {
-    const ready =
-      (this.entity[0] &&
-        !this.Graph.some(
-          element,
-          (index) =>
-            element._history === undefined &&
-            this.config.entities[index].show_graph !== false,
-        )) ||
-      this.config.show.loading_indicator !== true;
+    const ready = (this.entity[0] &&
+                    !this.Graph.some(
+                      element,
+                      (index) =>
+                        element._history === undefined &&
+                        this.config.entities[index].show_graph !== false,
+                    )
+                  )
+                  || this.config.show.loading_indicator !== true;
     return this.config.show.graph ? html`
       <div class="graph">
         ${ready ? html`
--- a/src/main.js
+++ b/src/main.js
@@ -324,8 +324,7 @@ class MiniGraphCard extends LitElement {
 
   renderGraph() {
     const ready = (this.entity[0] && !this.Graph.some(
-      element,
-      index => element._history === undefined && this.config.entities[index].show_graph !== false
+      (element, index) => element._history === undefined && this.config.entities[index].show_graph !== false,
     ))
     || this.config.show.loading_indicator !== true;
     return this.config.show.graph ? html`
--- a/src/main.js
+++ b/src/main.js
@@ -325,9 +325,9 @@ class MiniGraphCard extends LitElement {
   renderGraph() {
     const ready = (this.entity[0] && !this.Graph.some(
       element,
-      index => element._history === undefined && this.config.entities[index].show_graph !== false)
-      )
-      || this.config.show.loading_indicator !== true;
+      index => element._history === undefined && this.config.entities[index].show_graph !== false
+    ))
+    || this.config.show.loading_indicator !== true;
     return this.config.show.graph ? html`
       <div class="graph">
         ${ready ? html`
--- a/src/main.js
+++ b/src/main.js
@@ -323,8 +323,14 @@ class MiniGraphCard extends LitElement {
   }
 
   renderGraph() {
-    const ready = this.entity[0] && this.Graph[0]._history !== undefined;
-
+    let historyLoadingIncomplete = 0;
+    this.Graph.forEach((graph, index) => {
+      if (graph._history === undefined
+          && this.config.entities[index].show_graph !== false)
+        historyLoadingIncomplete += 1;
+    });
+    const ready = (this.entity[0] && (historyLoadingIncomplete === 0))
+                    || (this.config.show.loading_indicator !== true);
     return this.config.show.graph ? html`
       <div class="graph">
         ${ready ? html`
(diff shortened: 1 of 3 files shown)


src/main.js
```diff
@@ -323,8 +323,14 @@ class MiniGraphCard extends LitElement {
   }
 
   renderGraph() {
-    const ready = this.entity[0] && this.Graph[0]._history !== undefined;
```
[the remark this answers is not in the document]
ildar170975: > Why not simply

Because I know a little about JS )))
Please check my commit.
Note that I edited `element.history` to `element._history`.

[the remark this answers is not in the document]
ildar170975: As for the additional option: this "loading" thing works very slow on my iPhone 5S (((. So I proposed a way to disable it.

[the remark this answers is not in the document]
ildar170975: Keep getting an error "'element' not defined"...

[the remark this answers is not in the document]
akloeckner: Maybe the `element` would have been obvious for an expert. It wasn't obvious for me either. Sorry for messing this up. :-)

I had been told, efficient code was a goal of this project, originally. See https://github.com/kalkih/mini-graph-card/issues/1116#issuecomment-2313216516. So, maybe the loading indicator wasn't such a good idea after all? We could also resolve to changing it back to something static? Or hide it through card-mod?

I'm not saying we should actually do either of these things. I'm just thinking about options, currently. If your say, the option is best, I'll agree.

src/main.js
```diff
@@ -323,8 +323,11 @@ class MiniGraphCard extends LitElement {
...
-
+    const ready = (this.entity[0] && !this.Graph.some(
+      (element, index) => element._history === undefined
+      && this.config.entities[index].show_graph !== false,
+    ))
+    || this.config.show.loading_indicator !== true;
```
akloeckner, line 330:
```suggestion
    || this.config.show.loading_indicator === false:
```

Can we change this to avoid having to define the value in `show`? It might later bite our asses in the UI config. See https://github.com/kalkih/mini-graph-card/pull/1128#discussion_r1761831060.

ildar170975: > Sorry for messing this up. :-)

No problem, these small things are a part of doing anything. "Who does not do anything - does not make mistakes".

> had been told, efficient code was a goal of this project, originally.

Totally agree with the the main point that the original code is efficient. Currently we are adding small things; hopefully this "loading indicator" is a minor thing as well. I only noticed a slow "rotation" on an old iPhone 5S - and if we can disable this "loading indicator" by a new option - then it will not affect anything.

Mostly I like this indicator because it is a common thing in HA:
history page:
image

logbook:
image

Although we have this in history-graph:
image
We can choose this way of course - then we we'll have to load a local translation of this "Loading state history..." sentence (in "src\translations\xxxxxx.json").

ildar170975: > Can we change this to avoid having to define the value in show? It might later bite our asses in the UI config

1. Imho we should add this new option. Yes, it increases a complexity a bit - but then we will have a possibility to "lighten up" a look (i.e. w/o a "loading indicator").

2. As for the UI: honestly this is a less prioritized feature for me.

3. As for default value: agree we should avoid listing all options in `DEFAULT_SHOW` for simplicity.
So I will consider your proposal.
The expression `this.config.show.loading_indicator !== true` is `true` if the option is explicitly defined in user config as `false` or it is undefined (i.e. does not have a default value).
The expression `this.config.show.loading_indicator === false` is `true` if the option is explicitly defined in user config as `false`; if it is undefined - then the expression will be `false` (like a default value is `false`).
If we want the indicator to be displayed by default - then we should use my way `!== true`.
Otherwise - we should use your way `=== false`.
Please correct me if I made a mistake.

akloeckner: Let me think aloud, as this creates logical knots in my brain...

We're setting
```
ready = involved_logic || not loading_indicator
```

Which reduces, while loading, to
```
ready = not loading_indicator
```

The icon will be displayed, if `ready==false`. We want to show the indicator by default, so we want `not loading_indicator` to be `false` by default:

|  loading_indicator | ready | indicator shown |
| -- | -- | -- |
| false | true | no |
| true | false | yes |
| undefined | false | yes |

So, we should use `loading_indicator === false`. 

Right? 🤔 

I agree with everything else your wrote.
* Keep the blue circle.
* Add the option.
* Remove the explicit default setting.
* (And also reduced options in UI only)

ildar170975: Now I got knots myself)))
image
I honestly do not understand why `not undefined` = `false`...

Checked myself again, I made a mistake in my prev post:
image
Based on this - yes, we should use `loading_indicator === false`.
Also I will remove a default value from DEFAULT_SHOW .

akloeckner: Thanks, looks good! I'll test it when I have some minutes at home and merge it!

akloeckner:
I tested it and it looks good. So, ready for merge, except:

While testing, I came accross a thought: Do we want to show the loading indicator until the very last graph has been loaded? Or do we want to render partial data? Currently, we are preparing to do the former. The latter would simply require changing `some` to `every`, I guess. Until introduction of the loader the card rendered whatever data was there...

I don't think, it really makes a difference (on faster devices at least). It will be cleaner to wait until all data is there.

So, I think, we should keep the PR as is. Do you agree?

ildar170975:
As you said earlier, "I got knots in my brains" )).
Currently in this PR:
```
const ready = (this.entity[0] && !this.Graph.some(
      (element, index) => element._history === undefined
      && this.config.entities[index].show_graph !== false,
    )) ...
```
which I understand as "NOT some history is undefined" - i.e. ALL histories should be defined.
Means - "ready=false" & indicator ON until the last history is gathered.
(just a guess) Until the last graph is gathered, the card may change it's look (lower/upper bounds) - so may be we really should wait for the last history?

akloeckner:
:-) Ok. We're having the same knots... But I agree: Your current proposal is to wait until the last data has loaded. And it should stay that way. 👍

ildar170975:
Ok, let's merge it!

PR #1155, opened 2024-10-02
fix: height in grid

ildar170975:
Before:
image
```
  - type: grid
    columns: 2
    square: false
    title: not square
    cards:
      - &ref_graph_card
        type: custom:mini-graph-card
        entities:
          - entity: sensor.xiaomi_cg_1_co2
      - &ref_entity_card
        type: entity
        entity: sun.sun
      - *ref_entity_card
      - *ref_graph_card

  - type: grid
    columns: 2
    square: true
    title: square
    cards:
      - *ref_graph_card
      - *ref_entity_card
      - *ref_entity_card
      - *ref_graph_card
```
After:
image

Probably may be useful in case of using a new experimental "sections" views (honestly, do not care about this still beta feature).

PR #1199, opened 2025-01-13
fix: card height in grid

akloeckner:
This is my take on the sections/grid issue. :-)

This changes the card height to occupy the assigned height in grid/sections view. The Graph will be stretched to fill this space! That means, its aspect ration will change. Change the `height` option to make it look better. See this example:
image

I consider this better than simply fixing the card height, because the card remains with lots of empty space in this case. If we fill the space, the default look will be nicer. Here's how it looks, if only the card height is fixed:
image

Here's how it is done technically:

* :host.height = 100%
  => fixes height of entire card to fill space, if there is a defined space
* graph.flex = auto
  => assigns remaining space in card to the graph itself
* graph_container.height = 100%
  => makes the container fill the entire space
* graph__container__svg:
  .position=relative
  .width=100%
  .height=100%
  => makes the container use the entire space. Not only in flex width.
* The relative positioning changes the z-order of the graph and the labels. So, this is fixed in the main source code.
* The SVG also is freed aspect-ratio-wise, otherwise it could not fill the entire space.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ We recommend looking at the [Example usage section](#example-usage) to understan
 | cache | boolean | `true` | v0.9.0 | Enable/disable local caching of history data.
 | show | list |  | v0.2.0 | List of UI elements to display/hide, for available items see [available show options](#available-show-options).
 | animate | boolean | `false` | v0.2.0 | Add a reveal animation to the graph.
-| height | number | `150` | v0.0.1 | Set a custom height of the graph. Disregarded, if the card has a fixed size already, e.g. in sections/grid view.
+| height | number | `150` | v0.0.1 | Set a custom height of the graph. Disregarded, if the card has a fixed size already, e.g. in sections view or grid/horizontal-stack card.
 | bar_spacing | number | `4` | v0.9.0 | Set the spacing between bars in bar graph.
 | line_width | number | `5` | v0.0.1 | Set the thickness of the line.
 | line_color | string/list | `var(--accent-color)` | v0.0.1 | Set a custom color for the graph line, provide a list of colors for multiple graph entries.
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ We recommend looking at the [Example usage section](#example-usage) to understan
 | cache | boolean | `true` | v0.9.0 | Enable/disable local caching of history data.
 | show | list |  | v0.2.0 | List of UI elements to display/hide, for available items see [available show options](#available-show-options).
 | animate | boolean | `false` | v0.2.0 | Add a reveal animation to the graph.
-| height | number | `150` | v0.0.1 | Set a custom height of the line graph.
+| height | number | `150` | v0.0.1 | Set a custom height of the graph. Disregarded, if the card has a fixed size already, e.g. in sections/grid view.
 | bar_spacing | number | `4` | v0.9.0 | Set the spacing between bars in bar graph.
 | line_width | number | `5` | v0.0.1 | Set the thickness of the line.
 | line_color | string/list | `var(--accent-color)` | v0.0.1 | Set a custom color for the graph line, provide a list of colors for multiple graph entries.
--- a/src/main.js
+++ b/src/main.js
@@ -429,6 +429,7 @@ class MiniGraphCard extends LitElement {
         stroke-dasharray=${this.length[i] || 'none'} stroke-dashoffset=${this.length[i] || 'none'}
         stroke=${'white'}
         stroke-width=${this.config.line_width}
+        vector-effect='non-scaling-stroke'
         d=${this.line[i]}
       />`;
 
--- a/src/main.js
+++ b/src/main.js
@@ -429,7 +429,6 @@ class MiniGraphCard extends LitElement {
         stroke-dasharray=${this.length[i] || 'none'} stroke-dashoffset=${this.length[i] || 'none'}
         stroke=${'white'}
         stroke-width=${this.config.line_width}
-        vector-effect='non-scaling-stroke'
         d=${this.line[i]}
       />`;
 
@@ -443,16 +442,24 @@ class MiniGraphCard extends LitElement {
   renderSvgPoint(point, i) {
     const color = this.gradient[i] ? this.computeColor(point[V], i) : 'inherit';
     return svg`
-      <circle
-        class='line--point'
-        ?inactive=${this.tooltip.index !== point[3]}
-        style=${`--mcg-hover: ${color};`}
+    <g
+      class='line--point--group'
+      ?inactive=${this.tooltip.index !== point[3]}
+      @mouseover=${() => this.setTooltip(i, point[3], point[V])}
+      @mouseout=${() => (this.tooltip = {})}
+    >
+      <line
+        class='line--point--border'
+        x1=${point[X]} y1=${point[Y]} x2=${point[X]} y2=${point[Y]}
+        stroke-linecap="round" stroke-width=${this.config.line_width * 2}
         stroke=${color}
-        fill=${color}
-        cx=${point[X]} cy=${point[Y]} r=${this.config.line_width}
-        @mouseover=${() => this.setTooltip(i, point[3], point[V])}
-        @mouseout=${() => (this.tooltip = {})}
       />
+      <line
+        class='line--point--fill'
+        x1=${point[X]} y1=${point[Y]} x2=${point[X]} y2=${point[Y]}
+        stroke-linecap="round" stroke-width=${this.config.line_width}
+      />
+    </g>
     `;
   }
 
--- a/src/main.js
+++ b/src/main.js
@@ -329,11 +329,11 @@ class MiniGraphCard extends LitElement {
       <div class="graph">
         ${ready ? html`
             <div class="graph__container">
-              ${this.renderLabels()}
-              ${this.renderLabelsSecondary()}
               <div class="graph__container__svg">
                 ${this.renderSvg()}
               </div>
+              ${this.renderLabels()}
+              ${this.renderLabelsSecondary()}
             </div>
             ${this.renderLegend()}
         ` : html`
@@ -539,6 +539,7 @@ class MiniGraphCard extends LitElement {
     const { height } = this.config;
     return svg`
       <svg width='100%' height=${height !== 0 ? '100%' : 0} viewBox='0 0 500 ${height}'
+        preserveAspectRatio='none'
         @click=${e => e.stopPropagation()}>
         <g>
           <defs>
--- a/src/style.js
+++ b/src/style.js
@@ -232,6 +232,7 @@ const style = css`
   }
   .graph {
     flex: auto;
+    min-height: 0;
     align-self: flex-end;
     box-sizing: border-box;
     display: flex;
--- a/src/style.js
+++ b/src/style.js
@@ -4,6 +4,7 @@ const style = css`
   :host {
     display: flex;
     flex-direction: column;
+    height: 100%;
   }
   ha-card {
     flex-direction: column;
@@ -234,6 +235,7 @@ const style = css`
     right: 0;
   }
   .graph {
+    flex: auto;
     align-self: flex-end;
     box-sizing: border-box;
     display: flex;
@@ -245,10 +247,13 @@ const style = css`
     display: flex;
     flex-direction: row;
     position: relative;
+    height: 100%;
   }
   .graph__container__svg {
     cursor: default;
-    flex: 1;
+    position: relative;
+    width: 100%;
+    height: 100%;
   }
   svg {
     overflow: hidden;
--- a/src/style.js
+++ b/src/style.js
@@ -1,17 +1,13 @@
 import { css } from 'lit-element';
 
 const style = css`
-  :host {
-    display: flex;
-    flex-direction: column;
-    height: 100%;
-  }
   ha-card {
     flex-direction: column;
     flex: 1;
     padding: 16px 0;
     position: relative;
     overflow: hidden;
+    height: 100%;
   }
   ha-card > div {
     padding: 0px 16px 16px 16px;
--- a/src/style.js
+++ b/src/style.js
@@ -273,6 +273,22 @@ const style = css`
   .line[anim="false"] {
     animation: pop .25s cubic-bezier(0.215, 0.61, 0.355, 1) forwards;
   }
+  .line {
+    vector-effect: non-scaling-stroke;
+  }
+  .line--point--group {
+    cursor: pointer;
+  }
+  .line--point--border {
+    vector-effect: non-scaling-stroke;
+  }
+  .line--point--fill {
+    stroke: var(--primary-background-color, white);
+    vector-effect: non-scaling-stroke;
+  }
+  .line--point--group:hover .line--point--fill {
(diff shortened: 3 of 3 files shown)


src/style.js
```diff
@@ -4,6 +4,7 @@ const style = css`
   :host {
     display: flex;
     flex-direction: column;
+    height: 100%;
```
ildar170975, line 7: Check these examples:
image
```
type: horizontal-stack
title: stack
cards:
    name: default
    entities:
    name: height:100% for :host
    entities:
    card_mod:
      style: |
        :host {
          height: 100%;
        }
    name: my PR
    entities:
    card_mod:
      style: |
        :host {
          display: unset;
          flex-direction: unset;
        }
  - type: calendar
    entities:
      - calendar.calendar_1
```
```
type: grid
columns: 4
cards:
    name: default
    entities:
    name: height:100% for :host
    entities:
    card_mod:
      style: |
        :host {
          height: 100%;
        }
    name: my PR
    entities:
    card_mod:
      style: |
        :host {
          display: unset;
          flex-direction: unset;
        }
  - type: calendar
    entities:
      - calendar.calendar_1
title: grid
```
We need to remove these 3 lines:
```
    display: flex;
    flex-direction: column;
    height: 100%;
```

ildar170975, line 7: Check these examples:
image
1st card - default
2nd - added `height: 100%` for `:host`
3rd - my PR

```
type: horizontal-stack
title: stack
cards:
    name: default
    entities:
    name: height:100% for :host
    entities:
    card_mod:
      style: |
        :host {
          height: 100%;
        }
    name: my PR
    entities:
    card_mod:
      style: |
        :host {
          display: unset !important;
          flex-direction: unset !important;
        }
        ha-card {
          height: 100%;
        }
  - type: calendar
    entities:
      - calendar.calendar_1
```
```
type: grid
columns: 4
cards:
    name: default
    entities:
    name: height:100% for :host
    entities:
    card_mod:
      style: |
        :host {
          height: 100%;
        }
    name: my PR
    entities:
    card_mod:
      style: |
        :host {
          display: unset !important;
          flex-direction: unset !important;
        }
        ha-card {
          height: 100%;
        }
  - type: calendar
    entities:
      - calendar.calendar_1
title: grid
```

akloeckner: Yes, the 100% on host were not enough. But maybe the first 100% needs to go where you put it in your PR... I'll check this. This is what a horizontal stack looks like with a scaled graph:
image

ildar170975: Looks nice imho. Better than a tiny one. I will continue a review soon.
Also, we should check how a `height` option affects...

ildar170975: FYI: by default any card inside horiz stack gets `display: block`; inside `grid` - `display: inline` (automatically).

akloeckner: The height option sets the graph height in masonry. But if the height is already fixed by the view (grid, sections, horizontal stack), then the height option only affects the squishing of the stroke.

If we continue with this, we should have selvalt in the loop. They had some idea to counter the squishing. I am afraid, I diluted the discussion a bit now...

akloeckner: > FYI

Thanks... This does not make our life much easier... ;-)

ildar170975: > this does not make our life much easier

The thing is that we should not change `display` like it was done long time ago (that time grid card was not invented yet).
To keep a proper behaviour - we need to use default values (i.e. not overwrite them).

ildar170975: What I see is
image
because of `preserveAspectRatio='none'`.
SVG is elongated.
We need to think more...

ildar170975: We probably need to set this thing dynamically:
image

ildar170975: Note that if the `height` option is set (to 1000 for example) - it is accounted in that SVG size:
image
Means - we may LEAVE to a user to define how the card will look if stretched))))
Anyway, assume we found a way to stretch a graph properly. Then a user comes and asks "WTH height is not accounted?"
The funniest part: seems that in this case your PR will be same as my PR ))) (and it will be 3rd concurrent PR with same code for "height")

akloeckner: (Tagging @selvalt7 to have them in this discussion on the desired looks of the stretched card in grid/sections view.)

> because of preserveAspectRatio='none'.
SVG is elongated.

Yes. It can be fixed for the lines, as selvalt pointed out. Points are more difficult. But there is a suggestion in #1139, at which I couldn't look closer yet.

> We probably need to set this "0 0 500 100" thing dynamically

That is what selvalt proposed. But it recalculates the entire graph each time the window is resized. (Basically, because the number of rows is only available *after* the Graph is initialized.) I think it will be more natural and efficient to leave such resizing to CSS.

> we may LEAVE to a user to define how the card will look if stretched

That might make sense, actually... It would also make the change much easier. It will just not look pretty by default. :-/

> WTH 'height' option is not accounted?

It is not accounted correctly currently either: the width is set to 500 and the height to whatever is in the option. Then, the graph is scaled (preserving aspect ratio) to fit the card's width. This changes the height. By the way, this also changes line widths, which can be set in the options, but are thus not respected correctly either. So, these dimensions are a sensible first guess, in the best case. And completely useless in the worst case. :-P Based on this observation, I would even consider to remove this option... Then again, there would be no way to set a height in default masonry view.

> seems that in this case your PR will be same as my PR ))

I have no problem to close mine. I just wanted to create something focused and testable to discuss the stretched approach.

In summary, I think, we should 
* for default masonry view
  * at least have a way to set height in default view, so the option must stay.
  *  But maybe we should treat the height option differently: use it as actual height and only scale the graph in width.
  * This would require some de-squishing of the graph anyway. Which would, as a side-effect, ideally make the line width match what is set by the user.
* For sections view, 
  * we should disregard the height option and use available space. Because we have another height option (the number of rows) to respect, there is no need to respect contradicting options.
  * Alternatively, if this de-squishing becomes too hard, we should leave the height of the graph in section view up to the user and accept an initial ugly space in the card...
  * We should decide on the way to handle height in sections before we introduce the support for them. Otherwise people will start using it and the later fix becomes actually a breaking change...

So, I think, the questions are:
1.  Do we prefer to scale graphs in section view? From a simplicity perspective, I would say: yes. And I take @ildar170975's comment and @selvalt7's PR as a yes, too.
2. Do we succeed to make that happen? If we cannot, we have to fall back to the manual height setting.

I will thus have a look at the de-squishing of points @selvalt7 proposed and see how far I get with question no. 2.

But, please, let me know, if you think differently on the preferred way of sizing the card! That could save me some work. ;-)

selvalt7: IMHO, the graph should use the available space in the sections view. I see no point in using two options for sizing. Other views should use the height option, which could be used as a CSS property instead.
The solution for graph squishing so far looks promising from my initial tests.

akloeckner: Ok, so we seem to agree on this. I also checked the latest changes in #1139. I like the looks of the graph very much. Very neat fix for the line and point styles, @selvalt7! I also saw, you picked up the latest changes from #1155.

I therefore cherry-picked your points rendering and fixed up some remnant styles on the go. I also cherry-picked @ildar170975's latest change to the height of `:host` vs. `ha-card`.

So, I think, we now have a minimal set of changes required to properly support sections view. And it will be attributed to all of us. ;-)

Could you please test this PR and see, if we're missing any edge cases? I did some testing and all seems to work fine, but you had a deeper dive into these issues than I had...

I suggest then, we finish this PR here and then go on to discuss the other functionality in #1139 and possibly add it step by step. OK?

selvalt7: I have found one edge case, with sections, when graph clips outside the card with this config:
```yaml
show:
  extrema: true
layout_options:
  grid_columns: 4
  grid_rows: 4
```
I came up with an idea to set `min-height: 0` in [this selector](https://github.com/akloeckner/mini-graph-card/blob/3aabe5981456efd71b47e9e34b826add4b0ea9b0/src/style.js#L37) but also it could be done with size limiting in #1139.
Other than that, it seems fine to me.

akloeckner: Hmm. I played around a bit and there are lots of possibilities, how a user could make the card too small to fit everything in. If we make the card small enough, even the state will clip out of the bottom. 

I think, this cannot and should not really be avoided automatically in the general case. At least YAML power users might have legitimate reasons to do this. Then, they'll have to live with the card being cut off. Even this might make sense for some:
```yaml
layout_options:
  grid_columns: 2
  grid_rows: 2
height: 10
line_width: 1
```

However, the case you outlined should be fixed, because there is actually sufficient space available. The graph simply clips out of bound, because it has a default `height` set to 100px. And that is too tall. We should not change the default height, I think, because it would affect most of the users. We should not change it in section view only either, because this is only possible after initialization. So, I guess the CSS approach should be the way to go. 👍 

I'll try it tonight...

akloeckner:
I noticed another glitch, which I haven't been able to fix yet: Setting an absurd height option on the card (e.g. `height: 5000`) makes parts of the graph be clipped. See the top of this graph:
image

I believe it is linked to the mask inside the SVG. I wouldn't consider it a show-stopper. But if you have an idea how to fix it, we could do it already...

ildar170975:
What we have now:
The card is not properly placed in a cell in Grid card - does not occupy an available area.
This can be solved by a proper combination of `display` & `height` properties (like it was done already in all three PRs).

Next - the graph's DEFAULT height should not cause a "tiny graph in a high card" case.
This is what discussed in this current PR of @akloeckner.

Next - margins/paddings/fonts etc should adapt to a card's size.
This was discussed in PR of @selvalt7.

Guys, lets solve the task step-by-step:
1. Solve the minimal task with a card's height.
2. The graph may be tiny (if a card is "stretched up" vertically to adapt to a higher cell) - but a user may fix it by a proper `height`. Means - the default look may be bad, but a user can fix it.
3. Next - we add a nicely set default graph's height.
4. Next - we add a further customization if required.

I.e. - let's issue a release with at least 1 solved task.

README.md
```diff
@@ -92,7 +92,7 @@ We recommend looking at the [Example usage section](#example-usage) to understan
 | cache | boolean | `true` | v0.9.0 | Enable/disable local caching of history data.
 | show | list |  | v0.2.0 | List of UI elements to display/hide, for available items see [available show options](#available-show-options).
 | animate | boolean | `false` | v0.2.0 | Add a reveal animation to the graph.
-| height | number | `150` | v0.0.1 | Set a custom height of the line graph.
+| height | number | `150` | v0.0.1 | Set a custom height of the graph. Disregarded, if the card has a fixed size already, e.g. in sections/grid view.
```
ildar170975, line 95: "e.g. in sections view or grid/stack card"

src/style.js
ildar170975: Probably we should allow to set a height manually - but it should be limited: not more than a possible value.
Assume someone need to have a SMALLER graph? (Like for adding additional elements above a graph with card-mod)

akloeckner: (I was tempted to move this discussion into the main thread, but I realized, we're discussing here how advanced users might use the card. So, I'll keep it separate, in order for us to be able to close the discussion, once we're agreed.)

> Probably we should allow to set a height manually - but it should be limited: not more than a possible value.

That would mean, the documented behaviour would be:
* Graph will have actual height of `height` setting divided by 500, times the actual graph width. (Which is bit of legacy stupidity already.)
* In sections view (or any other location that fixes a total card height), the actual height of the graph will be limited to the available space.
* If you want the graph to fill all the available space, set a suitable `height` setting well above the imaginable available space.

While I think, that should be possible by setting a flex option which doesn't allow growing (but only shrinking)... I think, this is quite un-intuitive... A "normal" (as opposed to "advanced") user would expect the graph to just look nice without setting an arbitrary but high enough `height`. Wouldn't they?

> Assume someone need to have a SMALLER graph? (Like for adding additional elements above a graph with card-mod)

I don't think we should design the card to be as hackable as possible. At least not, if it impairs the "normal" user's experience. I think, if people have such specific needs for modifications, they will find a way, regardless of what we do... For example, they could simply override the flex setting themselves.

akloeckner:
> 2. The graph may be tiny (if a card is "stretched up" vertically to adapt to a higher cell) - but a user may fix it by a proper height. Means - the default look may be bad, but a user can fix it.

I don't think, we should release something like this. Because users will change the heights of their graphs. And then we come and change the card's behaviour, again! If we know, it is not desirable, we should avoid it. That's why I would prefer to solve this step together with the first step. That way, we provide something which works out of the box for the "normal" users. If then some power users asks for an interface to hack the card, we can consider it.

> Next - margins/paddings/fonts etc should adapt to a card's size.

This should be discussed later, I agree. 

To be honest: I think, the solution we merged together from our three branches here is basically the solution to your steps 1, 2 and 3 already. Without an "ugly" intermediate step. Unless I am missing something (other than the open discussion on the card-mod hackability). So, I would suggest to take these three steps. And then think about the next. If you're concerned about traceability, we can merge your PR first and base this one on yours. But I would not create an intermediate release.

README.md
```diff
@@ -92,7 +92,7 @@ We recommend looking at the [Example usage section](#example-usage) to understan
 | cache | boolean | `true` | v0.9.0 | Enable/disable local caching of history data.
 | show | list |  | v0.2.0 | List of UI elements to display/hide, for available items see [available show options](#available-show-options).
 | animate | boolean | `false` | v0.2.0 | Add a reveal animation to the graph.
-| height | number | `150` | v0.0.1 | Set a custom height of the line graph.
+| height | number | `150` | v0.0.1 | Set a custom height of the graph. Disregarded, if the card has a fixed size already, e.g. in sections view or grid/horizontal-stack card.
```
ildar170975, line 95:
```suggestion
| height | number | `150` | v0.0.1 | Set a custom height of the graph. Disregarded if the card has a fixed size already, e.g. in sections view or grid/horizontal-stack card.
```

src/main.js
```diff
@@ -539,6 +547,7 @@ class MiniGraphCard extends LitElement {
     const { height } = this.config;
     return svg`
       <svg width='100%' height=${height !== 0 ? '100%' : 0} viewBox='0 0 500 ${height}'
+        preserveAspectRatio='none'
         @click=${e => e.stopPropagation()}>
```
Liquidmasl, line 551: the readme suggests that height has no effect in grid view or section view, but this does not seam to be the case. The height is still used for the viewbox. Which is mostly fine, but for very small values this breaks the graph. (everything below 10). 

So in section view this should probably be just set to something sensible.

Or implement the function I have seen in the other PR that tackles the same topic, even though it wasnt wired in over there as well, it would solve that issue: `getGraphHeightSections()` -- https://github.com/kalkih/mini-graph-card/pull/1139/changes#diff-27653c212e1cfe533e4eb2f7d0d3f89604c9de48a09583b4cbbbcbd08a07da79R153

README.md
Liquidmasl, line 95: its not disregarded atm. it sused in the svg viewbox still, I left a review comment over there

Liquidmasl (changes requested):
This is a bit cleaner implemented then https://github.com/kalkih/mini-graph-card/pull/1139

But it should implement the HA API `getLayoutOptions()` so HA doesnt have to guess the initial card size. 
Other then that and the svg viewport height think and slightly inaccurate readme, this looks good. I would really like to get this merged

kattcrazy:
Any progress on this? :)

Issue #630, opened 2021-07-14
[Feature request] Negative values


roybosch:
Hi!

First of all: great lovelace card. I'm using the mini-graph-card for all my graphs because it's by far the best out there.

Since a few weeks I'm also using it to monitor my energy usage compared to my solar yield. I made a template sensor which subtracts one from another, resulting in a positive value (I'm using more energy than my solar panels can deliver) or a negative value (the yield of my solar panels is more than I'm using). With the color_threshold attribute I'm making these two states red or green as shown below:

![image](https://user-images.githubusercontent.com/45666392/125592957-4679dd7c-e5ce-48f3-a749-99e430871430.png)

There could be a few improvements on this, which aren't possible so far:

1. I'd like to draw an x-axis to show where the value goes negative. With the green and red color I can make this also visible, but showing a line would communicate better I think. Showing a small timeline on the x-axis would make it even better.
2. If I change the graph to **bars** instead of **line**, the negative bars are drawn in positive direction. Only the color changes. I'd like to see them move down from the x-axis to state that these are negative values. Like this:
![image](https://user-images.githubusercontent.com/45666392/125595439-686970c6-0a4b-43b6-8459-e2b55ff8e2ee.png)

3. If I'm using the **show_fill: true** attribute on the graph shown above, it colors the area under the line. It would be nice if it colors the negative values above the line until the x-axis. Like so:
![image](https://user-images.githubusercontent.com/45666392/125594547-953c6ef5-aba5-4eaf-a628-76db3c12a7df.png)

Just some ideas which make your card even more powerfull in my opinion!

Kind regards,

Roy

ildar170975:
This could be a very nice feature.

trankillity:
[Example SVG implementation](https://github.com/kalkih/mini-graph-card/issues/418#issuecomment-672405667) in my feature request from almost 2 years ago.

mareklab:
<img width="461" alt="image" src="https://user-images.githubusercontent.com/10220070/182361166-76788d5f-7d74-480a-8a03-e7e4df24ceae.png">
Yes, this could be very useful to display above as bars instead of ugly lines. 
Home Assistant Energy page got it right:
<img width="1092" alt="image" src="https://user-images.githubusercontent.com/10220070/182361757-9826c88c-1d96-493b-80cd-7754a495158f.png">
 just it cannot be reused outside of Energy :(

trankillity:
[...]

PR #1205, opened 2025-01-28
feat: add Fill Threshold feature - making graph filling height customizable

Haluska77:
This change makes graph filling height customizable. 

Currently, each line (sensor) fills whole height of the graph from its value to the bottom of the graph like on the picture
mini-graph-before

Blue line represents negative values, purple color represents positive values. Picture graph shows solar production (yellow) and grid power (blue = from grid, purple = to grid).
The amount of consumed energy from grid (blue) is not correctly displayed as it should be filled from zero to its value.
It should look like this:
mini-graph-after

This change will fix it. It adds a new config variable 'fill_threshold' which represents user expected value that graph is going to be filled up. 'fill_threshold' is optional. If 'fill_threshold' contains string, empty string or any non number value, then it is ignored.


akloeckner:
Thanks, @Haluska77! This looks like a useful feature. I somehow cannot really imagine, how it works. I'm looking at your screenshot and something doesn't fall in place in my head. (Although the change looks very simple.) Could you provide an example configuration, please?

Haluska77:
hi, glad to hear you like it. This particular picture above is based on following configuration:
```
                entities:
                  - entity: sensor.grid_power_negative
                  - entity: sensor.solar_production
                    color: "#ffee3A"
                color_thresholds:
                  - value: -10000
                    color: "#488fc2"
                  - value: 0
                    color: "#8353d1"
                fill_threshold: 0
                hours_to_show: 12
                line_width: 2
                show:
                  state: false
                  name: false
                  icon: false
                  legend: false
                  zero_line: true
                points_per_hour: 4

```

ildar170975:
Interesting. Will test.

ildar170975:
A quick question.
There is a new `fill_threshold` option.
Assume it does smth in the proposed algorithm.
Imho in the ideal scenario ALL graphs with negative values should be filled "from a curve to Y=0" (or to min(abs(lower_bound), abs(upper_bound))). Means - none option is needed. Then why was this new `fill_threshold` option proposed?
Sorry if the question is silly.

Haluska77:
I extended solution for each entity
mini-graph-fully-custom

In this case, each entity could be configurable, which heigh to be filled.

This could be reached by 
```
                  - entity: sensor.grid_power_negative
                    fill_threshold: 0
                  - entity: sensor.solar_production
                    fill_threshold: -3000
                    color: "#ffee3A"

```
This could be combined with 'fill_threshold' on 'entities' level, but 'entity' level overides value on 'entities' (parent) level.

So following configuration do the same result:
```
                entities:
                  - entity: sensor.grid_power_negative
                  - entity: sensor.solar_production
                    fill_threshold: -3000
                    color: "#ffee3A"
                fill_threshold: 0

```

Haluska77:
> A quick question. There is a new `fill_threshold` option. Assume it does smth in the proposed algorithm. Imho in the ideal scenario ALL graphs with negative values (or positive and negative values) should be filled "from a curve to Y=0" (or to min(abs(lower_bound), abs(upper_bound))). Means - none option is needed. Then why was this new `fill_threshold` option proposed? Sorry if the question is silly.

'fill_threshold' is the value, it might not be only 0, but could be setup on your own. 
Maybe an additional extension is answer to your question when you are able to setup threshold for specific graph (line) you want, not all.

ildar170975:
We need to add an X-axis line ASAP for this...

Update 19.06.26: proposed in https://github.com/kalkih/mini-graph-card/pull/1378

akloeckner:
> We need to add an X-axis line ASAP for this...

Could it be done by card-mod-styling the line of the fill? I think, it would create a horizontal line.

ildar170975:
> Could it be done by card-mod-styling the line of the fill?

Could be, but failed to get a result in Safari.
And it works only when a "fill" starts not from a bottom, but from a 0.

akloeckner:
> Could be, but failed to get a result in Safari.

Yes, I was thinking too simply. The fill is not just a fill. It's a filled (white) mask, which is then applied to the gradient rect following it. So, adding a stroke to the mask simply increases the size of the fill. 🤷

ildar170975:
> If we are thinking about adding a zero line

I wonder if it is possible to add a zero line "simply" as an svg element like we added (incompletely) grid lines?
Also, think about a possibility to add SEVERAL static lines (there is a FR in repo) with properties like "y_value, y_axis, line_color, line_style, line_width" - then this zero line could be just some of them.
Sorry for an offtopic.

akloeckner:
> Sorry for an offtopic.

No, it's good. And I don't think it's off topic per se. But I agree, let's consider it off topic and not add the line in this PR. But maybe as part of the grid-line PR. 👍

ildar170975:
> But maybe as part of the grid-line PR. 👍

I would disagree. Vertical grid lines depend on an time span, fixed horizontal lines' position depend on bounds. So let's propose this functionality in a separate PR.
Meanwhile I am going to test THIS proposed functionality (sorry for delays, I have rather limited resources these days).

wbu:
Looks like this fixes #630, right?

jagsta:
Is this planned to be merged into main? This is functionality i'd really love to be able to use.

ildar170975:
> this planned to be merged into main?

It is planned to analyze this PR with a more attention. 
Update: we have some number of PRs to be jointly approved and then merged; hopefully it will be a series of dev releases. Please be patient.

ildar170975:
@Haluska77 
Hello!
1. Added some changes:
-- descriptions for `getFill()` & `_calcY()` functions (the latter was used in the `getFill()` , so it is involved somehow);
-- refactored to use a `isNumeric()` helper.
2. Problem - the PR does not account a `fill: fade` option: a fade is processed for a WHOLE fill, not separately to an "above" & "below" parts. Not a big deal though, can be improved in a future (update 08.09.26 - fixed in https://github.com/kalkih/mini-graph-card/pull/1443); meanwhile we can note in Docs about this limitation.
изображение

3. Please add proposed changes for Readme for the added `fill_threshold` options (global & per_entity). If you will not, will do it myself in a separate PR...

I am testing this PR, will take some time.

ildar170975:
Also, I think we need to rename the option to “baseline”, ~I am going to extend it to bars graphs~.

ildar170975:
More required fixes & updated Docs can be added in new PRs (like in https://github.com/kalkih/mini-graph-card/pull/1408).

PR #1217, opened 2025-02-25
fix: Use a proper config in renderState()

### LINKED ISSUE #698: "state_adaptive_color" option is confusing since the left "state" text is used for displaying states for all entities

ildar170975:
Fixes https://github.com/kalkih/mini-graph-card/issues/698#issuecomment-2661560913

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -291,14 +291,16 @@ class MiniGraphCard extends LitElement {
       const state = this.getEntityState(id);
       // use tooltip data for main state element, if tooltip is active
       const { entity: tooltipEntity, value: tooltipValue } = this.tooltip;
-      const value = isPrimary && tooltipEntity !== undefined ? tooltipValue : state;
-      const entity = isPrimary && tooltipEntity !== undefined ? tooltipEntity : id;
+      const condition = isPrimary && tooltipEntity !== undefined;
+      const value = condition ? tooltipValue : state;
+      const entity = condition ? tooltipEntity : id;
+      const entity_config = condition ? this.config.entities[entity] : entityConfig;
       return html`
         <div
           class="state ${!isPrimary && 'state--small'}"
           @click=${e => this.handlePopup(e, this.entity[id])}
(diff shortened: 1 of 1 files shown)


akloeckner (commented):
Thanks! I had thought about starting this already myself, but then delayed it, because I wasn't sure about the logics of the changes. I left some recommendations...

src/main.js
```diff
@@ -259,12 +259,11 @@ class MiniGraphCard extends LitElement {
...
 
   renderStates() {
-    const [firstEntityConfig] = this.config.entities;
     if (this.config.show.state)
       return html`
         <div class="states flex" loc=${this.config.align_state}>
-          ${this.renderState(firstEntityConfig, 0)}
-          <div class="states--secondary">${this.config.entities.map((entityConfig, i) => i > 0 && this.renderState(entityConfig, i) || '')}</div>
+          ${this.renderState(0)}
+          <div class="states--secondary">${i => i > 0 && this.renderState(i) || ''}</div>
```
akloeckner, line 266: I think, this lost something: We should be iterating over all `this.config.entities` and only rendering the state, if `i > 0` (because we have rendered the first state already).

akloeckner (changes requested):

akloeckner (approved):

PR #1235, opened 2025-04-29
fix: computeName() when no "friendly_name" available

ildar170975:
Currently for numerical input helpers we see this:
image

No name is shown for the `input_number` entity (both on a top of the card & in a legend).

This happens because a `name` option was not defined for an entity & a `friendly_name` was not explicitly set as well.

With this PR:
image

Test code:
```
type: custom:mini-graph-card
entities:
  - entity: sensor.some_sensor
  - entity: input_number.testing_number
```

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -691,7 +691,9 @@ class MiniGraphCard extends LitElement {
   }
 
   computeName(index) {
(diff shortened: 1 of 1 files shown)


akloeckner (approved):
Nice catch...

PR #1236, opened 2025-04-29
fix: computeUom() for attribute & empty unit

ildar170975:
Two issues are fixed:

1. When an `attribute` option is defined & no `unit` is set - ignore `unit_of_measurement`.

Entity:
image

Before:
image

After:
image

Test card:
```
type: custom:mini-graph-card
entities:
  - entity: sensor.carbon_dioxide
    attribute: battery_level
```

Actually, HA frontend has a support of uoms for attributes; this hopefully will be implemented in `mini-graph-card` in future.

2. When an empty unit is defined - a default uom is still used:
Before:
image

After:
image

Test cards:
```
        name: empty for entity
        entities:
          - entity: sensor.carbon_dioxide
            unit: ''

        name: empty for card
        entities:
          - entity: sensor.carbon_dioxide
        unit: ''
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ We recommend looking at the [Example usage section](#example-usage) to understan
 | icon | string |  | v0.0.1 | Set a custom icon from any of the available mdi icons.
 | icon_image | string |  | v0.12.0 | Override icon with an image url
 | name | string |  | v0.0.1 | Set a custom name which is displayed beside the icon.
-| unit | string |  | v0.0.1 | Set a custom unit of measurement.
+| unit | string |  | v0.0.1 | Set a custom unit of measurement (`''` value for an empty unit).
 | tap_action | [action object](#action-object-options) |  | v0.7.0 | Action on click/tap.
 | group | boolean | `false` | v0.2.0 | Disable paddings and box-shadow, useful when nesting the card.
 | hours_to_show | integer | `24` | v0.0.2 | Specify how many hours of history the graph should present.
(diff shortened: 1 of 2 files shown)


akloeckner (approved):
This would be a perfect use case for the [Nullish_coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) operator. But as long as we're not on the most recent JavaScript version, we can't use it... 🤷

ildar170975:
> This would be a perfect use case for the [Nullish_coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) operator.


Also wanted to use it!!!! But it is supported in TS, not in some JS versions.

akloeckner:
We could bump the ECMA script version used in Babel to something more recent. And then the operator would be there. I have tried that, but it will require updating lots of dependencies. Which is also a step we should go at some point in time. But it failed so badly on me, that I gave up for now.

ildar170975:
> will require updating lots of dependencies

Agree, we need to do it. I see some people do it regularly... Have no knowledge to do it myself, unfortunately... Is there any automatic process for that?

akloeckner:
> automatic process

I think so, but I have never used it myself. 

I guess, the first huge step would be to catch up manually. From there on, the changes will be less. 

However, my being inactive here recently tells me to not take on too much of a challenge, currently...

Issue #555, opened 2021-02-05
Feature Request: individual entity line styles


RoboMagus:
Would be nice to have all graph styling options available as a per entity configuration. E.g. line_width and color_thresholds.
Additionally, it would be a great addition to have additional line styles other than a solid line. For instance a dashed / dotted line.

RomRider:
Thanks for this feature request.

This feature is already available in [apexcharts-card](https://RomRider/apexcharts-card).
This card is probably going to stay in maintenance mode, so no new features will be added in the future.

LoneWolf345:
I would second that this feature should be brought over to mini-graph-card.

ildar170975:
check v0.14

PR #1241, opened 2025-04-29
feat: line_width per entity

ildar170975:
Partly implements https://github.com/kalkih/mini-graph-card/issues/555

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,7 @@ properties of the Entity object detailed in the following table (as per `sensor.
 | entity ***(required)*** | string |         | Entity id of the sensor.
 | attribute | string |         | Retrieves an attribute or [sub-attribute (attr1.attr2...)](#accessing-attributes-in-complex-structures) instead of the state
 | name | string |         | Set a custom display name, defaults to entity's friendly_name.
+| line_width | number |         | Override for a thickness of the line.
 | color | string |         | Set a custom color, overrides all other color options including thresholds.
 | unit | string |         | Set a custom unit of measurement, overrides `unit` set in base config (`''` value for an empty unit).
 | aggregate_func | string |         | Override for aggregate function used to calculate point on the graph, `avg`, `median`, `min`, `max`, `first`, `last`, `sum`.
--- a/README.md
+++ b/README.md
@@ -134,7 +134,7 @@ properties of the Entity object detailed in the following table (as per `sensor.
 | aggregate_func | string |         | Override for aggregate function used to calculate point on the graph, `avg`, `median`, `min`, `max`, `first`, `last`, `sum`.
 | show_state | boolean |         | Display the current state.
 | show_legend_state | boolean |  false  | Display the current state as part of the legend.
-| show_indicator | boolean |         | Display a color indicator next to the state, (only when more than two states are visible).
+| show_indicator | boolean |         | Display a color indicator next to the state.
 | show_graph | boolean |         | Set to false to completely hide the entity in the graph.
 | show_line | boolean |         | Set to false to hide the line.
 | show_fill | boolean |         | Set to false to hide the fill.
--- a/src/buildConfig.js
+++ b/src/buildConfig.js
@@ -6,6 +6,7 @@ import {
   MAX_BARS,
   DEFAULT_COLORS,
   DEFAULT_SHOW,
+  DEFAULT_MARGIN,
 } from './const';
 
 /**
@@ -122,7 +123,7 @@ export default (config) => {
     line_color: [...DEFAULT_COLORS],
     color_thresholds: [],
     color_thresholds_transition: 'smooth',
-    line_width: 5,
+    line_width: DEFAULT_MARGIN,
     bar_spacing: 4,
     compress: true,
     smoothing: true,
--- a/src/const.js
+++ b/src/const.js
@@ -2,6 +2,7 @@ const URL_DOCS = 'https://github.com/kalkih/mini-graph-card/blob/master/README.m
 const FONT_SIZE = 14;
 const FONT_SIZE_HEADER = 14;
 const MAX_BARS = 96;
+const DEFAULT_MARGIN = 5;
 const ICONS = {
   humidity: 'hass:water-percent',
   illuminance: 'hass:brightness-5',
@@ -55,6 +56,7 @@ export {
   FONT_SIZE,
   FONT_SIZE_HEADER,
   MAX_BARS,
+  DEFAULT_MARGIN,
   ICONS,
   DEFAULT_COLORS,
   UPDATE_PROPS,
--- a/src/main.js
+++ b/src/main.js
@@ -103,7 +103,7 @@ class MiniGraphCard extends LitElement {
   getMaxLineWidth() {
     const arr = this.config.entities
       .map(entityConfig => entityConfig.line_width)
-      .filter(line_width => !Number.isNaN(line_width));
+      .filter(line_width => line_width !== undefined && !Number.isNaN(line_width));
     return Math.max(
       this.config.line_width,
       ...arr,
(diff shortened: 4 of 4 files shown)


ZionFox:
@kalkih This would be a wonderful feature to have.

ildar170975:
ping @akloeckner

fran68:
I think the function getMinMaxLineWidth wouldn't meet the following use cases:
- line_width is set for entity but graph is not shown (show_graph: false). line_width is evaluated although it shouldn't
- if line_width is set per entity for all entities, the global line_width should be out of scope

Maybe you can a take a look at #[e0d0323](https://github.com/fran68/mini-graph-card-xt/commit/e0d03233028d195f3a174d6a9eb342271ca1a0fb)

And a question: Why is min_line_width used for x margin instead of max_line_width? Wouldn't that clip lines with higher line_width at the edge.

ildar170975:
> I think the function getMinMaxLineWidth wouldn't meet the following use cases:
> 
>     * line_width is set for entity but graph is not shown (show_graph: false). line_width is evaluated although it shouldn't
> 
>     * if line_width is set per entity for all entities, the global line_width should be out of scope
> 
> 
> Maybe you can a take a look at #[e0d0323](https://github.com/fran68/mini-graph-card-xt/commit/e0d03233028d195f3a174d6a9eb342271ca1a0fb)

I honestly really like your idea!
https://github.com/fran68/mini-graph-card-xt/commit/e0d03233028d195f3a174d6a9eb342271ca1a0fb#diff-3274f1a37032fb0ae4e2823def0007c634e869ae0dfc304ff6a12c36513c3a52R13
But - I am not sure we need this `getMinMaxNumber()` as a separate function.
The function (abstract from a particular option) takes an `arr_obj` array of objects (dicts), and either uses a "local" `item[opt]` option value or a "global" option value.
Great, but probably the `line_width` case could be the ONLY case where this function is needed as a separate function.

Meanwhile, I added a few lines to implement your cases "show_graph: false" & "every entity has line_width" - took your code as an inspiration.
Please check the added [commit](https://github.com/kalkih/mini-graph-card/pull/1241/commits/3e3c95966488fd50ba7cb8d74896e56e0567a250) (do not bother about linter errors, they are unrelated & fixed in the next commit).
May be a bit more cumbersome than your separate function though...

ildar170975:
> And a question: Why is min_line_width used for x margin instead of max_line_width? Wouldn't that clip lines with higher line_width at the edge with higher line_width.

Need to check once again...

Update:
Since it was decided some time ago, could not recall a reason.
But now I recall: the reason was that I did not like a horizontal spacing between an Y-axis and a curve.
Compare this:

margin[X] = min, margin[Y] = max:
image

margin[X] = max, margin[Y] = max - see a horizontal gap:
image

So, I chose the 1st variant.

ildar170975:
@fran68 
Also, (might want to test in your setup), no need to calculate min & max values for bars.
https://github.com/kalkih/mini-graph-card/pull/1241/commits/8105211901e3395a22a4ba856e07e245b5861885

fran68:
> @fran68 Also, (might want to test in your setup), no need to calculate min & max values for bars. [8105211](https://github.com/kalkih/mini-graph-card/commit/8105211901e3395a22a4ba856e07e245b5861885)

Wouldn't it be better to place it in the filter?

fran68:
Thanks for explanation! 

However, isn't the design approach to have a line which has round shapes at the ends and does not touch the edge of the diagram when fill is set to false?

I would prefer the max option, maybe by extending the lines. Or we handle it in the way when the fill is set to true/fade, no gap at all!?

ildar170975:
> Wouldn't it be better to place it in the filter?

Probably better, cannot say, away from a PC, will be able to check in a week I hope. 

Update 06.06.26:
Sorry for a delay with reply!
I have not figured out how to place it in to a filter.
We have an array of `entities`, first filtering out entities with a shown graph (`show_graph !== false`), then saving either a valid per-entity `line_width` or a global `line_width`. My thought was that this operation should not be performed at all in case of bars.
Tested a bit with v.0.13 and a simple setup:
```
type: custom:mini-graph-card
entities:
  - entity: sensor.xxx
  - entity: sensor.yyy
  - entity: sensor.zzz
show:
  graph: bar
  fill: false
line_width: 20
```
and it gives an unexpected right margin:
изображение
I.e. in case of `fill: false` & bars a glitch occurs.
Need some time to fix that.
Imho all margins should be processed differently in case of bars.

ildar170975:
> Thanks for explanation!

Sorry, cannot answer now. For these kind of questions , I need a PC nearby (((
Just a quick reply: we may start with the proposed algorithm, and then start playing, then improve if needed. 


Update 06.06.26:
> However, isn't the design approach to have a line which has round shapes at the ends and does not touch the edge of the diagram when fill is set to false?
> I would prefer the max option, maybe by extending the lines. Or we handle it in the way when the fill is set to true/fade, no gap at all!?

Check this line:
https://github.com/kalkih/mini-graph-card/blob/e10c2e7b53e0a450ba2fa0adecccd026782bb2ce/src/main.js#L114
When `fill: true/fade` -> margin=0, otherwise it is `this.config.line_width, this.config.line_width`.
This gives us with untouched v.0.13.0 & this simple example:
```
type: custom:mini-graph-card
entities:
  - entity: sensor.xxx
  - entity: sensor.yyy
  - entity: sensor.zzz
show:
  fill: false
  points: true
```
this picture:
изображение

zoomed in:
изображение

when a point is shown:
изображение

i.e. a point does touch an Y-axis (even partly intersect it for 1/2 of stroke-width = 1/4 of line_width), a line (when no points are shown) - does not touch.

Then, what to do when we have different per-entity `line_width`?
(below - tests with a code from this PR):
```
type: custom:mini-graph-card
entities:
  - entity: sensor.xxx
    line_width: 15
  - entity: sensor.yyy
    line_width: 10
  - entity: sensor.zzz
show:
  fill: false
  points: true
line_width: 5 // i.e. default
```
w/o points:
изображение

with points:
изображение

Same presentation if we compare the curves with a MINIMAL `line_width`.
Now imagine what happens if we use a MAXIMUM `line_width` (i.e. max option) - a horizontal gap between a curve with a minimal `line_width` will be much larger ([see here](https://github.com/kalkih/mini-graph-card/pull/1241#issuecomment-4526955162)).

ildar170975:
@fran68 
Provided more detailed answers, sorry for a delay.

fran68:
> I.e. in case of `fill: false` & bars a glitch occurs. Need some time to fix that. Imho all margins should be processed differently in case of bars.

The same topic I was referring to (1. in https://github.com/kalkih/mini-graph-card/pull/1363#issuecomment-4579955991 )

fran68:
> 
> Check this line:
> 
> https://github.com/kalkih/mini-graph-card/blob/e10c2e7b53e0a450ba2fa0adecccd026782bb2ce/src/main.js#L114
> 
> 
That is my point. I think the design decision was to have different looks for the two line types (with fill and without fill). In case of fill set to true/fade there is no gap between line end and the diagram border. Sure, to have a completly filled area between line, border and bottom of diagram. In case of fill set to false I would say the goal was to have a free-floating line with round ends not touching the border. 

The latter is fine for lines with same line_width. With different line widths there is a challenge using the right line width:

- using the minimum line with:

You're right the gap is smaller but the entities/lines with higher line width values might touch the diagram border. In my opinion it breaks the look if my assumption above is right regarding the design approach. Then there is no need to keep a x-margin at all.

- using the maximum line width:

The lines might not touch the diagramm border. But for higher line_widths there are larger gaps as you've written .

My focus above is only on lines w/o points. I decided for me to use maximum line width, keep the free floating lines and extend the lines by adapting the code in getPath.

mcg_line_width

ildar170975:
Again away from PC for a few days. 

For margin problem with bars: yes, same problem which you highlighted. Will try to fix it when be near PC.
I think this should be done in a separate PR, may be even not in your PR for bar spacing, imho this is a deeper issue since it occurs in a current v.0.13. 

For line_width problem: to test the algorithm, I used big values of line_width (5, 10, 15). But in real life probably people will not use such big values. In this case - may be a difference between “use max or min” will be less significant. 

In this PR , I will keep using “min”, let us see if people like or dislike it…

Please keep in touch, your advices and observations are valuable.

PR #1247, opened 2025-05-12
fix: deprecated paper-item-icon-color

ildar170975:
paper-item-icon-color is deprecated in 2025.5


ildar170975:
@akloeckner 
This is a fix which should be added in the forthcoming release.
As for the "loading indicator" thing - will issue a fix in a week when be at home.

akloeckner (approved):

PR #1253, opened 2025-05-27
docs: Add "My Home Assistant" HACS Link to README

DJaeger:
To make it easier to install the card


akloeckner (approved):
Good idea. I changed the target branch to dev, so we get it in the regular workflow. We're planning to release on master in the nearer future, so it should end up on the main page in a short while.

PR #1270, opened 2025-06-25
fix: extend d3-interpolateRgb() to process CSS vars

### LINKED ISSUE #1259: Colors defined by CSS vars in "color_thresholds" are wrongly processed
Hey everyone.

On 0.13.0 color vars `color: var(---color-var)` renders black bars on the card.
If i change to `color: green` it works ok.

Just rolled back to check, and this example code works ok on 0.12.1:
```
            name: Master CPU Temp
            icon: mdi:temperature-celsius
            entities:
              - entity: sensor.processor_temperature
            show:
              graph: bar
            hours_to_show: 24
            points_per_hour: 3
            color_thresholds:
              - value: 0
                color: var(--success-color)
              - value: 40
                color: var(--error-color)
              - value: 65
                color: var(--warning-color)
```

Hope my submission includes all needed info.
Let me know if you need any further information from me.

ildar170975:
Seems that an external d3-interpolateRgb() does not process CSS variables like `var(--green-color)`.
The d3-interpolateRgb() function is replaced with a proxy function which converts CSS variables to values.
Will hopefully solve https://github.com/kalkih/mini-graph-card/issues/1259.


akloeckner:
I'm not sure, if we should do this. When I added d3, using a hidden element was an option for interpolation, too. I explicitly chose to not do it. I don't fully recall the reasons. I think, it was somehow considered unstable. And even after adding the small d3 library, it was criticized for introducing a dependency.

So, I think our options are:
* Choose to not support css variables.
* Switch the entire interpolation to css hidden elements. That way we would support the full css language. Because the next thing will be someone requesting to use mixed colors or the likes.

ildar170975:
Thanks a lot for the feedback!!! Converting to draft, will think about it later.

ildar170975:
> * Switch the entire interpolation to css hidden elements

This is unclear for me, will check your previous posts to understand the point.

> * Because the next thing will be someone requesting to use mixed colors or the likes.

BTW, `interpolateRgb` first converts a color before processing, different formats seem to be supported.
https://github.com/d3/d3-color/blob/71c7f100f9fa85a1c70fcbaeb5f803ee8db5620d/src/color.js#L201

> criticized for introducing a dependency.

I introduced a separate file "interpolateRgb.js" exactly to REPLACE a dependency in some future.

akloeckner:
> Switch the entire interpolation to css hidden elements

We could also *set* some colour property on the `element`, using mix-ins. And then retrieving that colour property will return the interpolated value as RGB. See the second example here: https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/getPropertyValue#examples .

Or we could declare the colour of each bar not as hex directly, but as mix-in string. 

The first variant is the somehow unstable one I tried to avoid. The second variant might be elegant, because we could get rid of the JS interpolation altogether.

akloeckner:
PS, looking at the issue now: https://github.com/kalkih/mini-graph-card/issues/1259#issuecomment-3001711060

The graph itself does render the variables correctly, because it uses the colours in a map directly. So it basically relies on the browser for interpolating. The points, however, and bars are interpolated in JS.

ildar170975:
OK, some info to be processed...

ildar170975:
> The graph itself does render the variables correctly, because it uses the colours in a map directly. So it basically relies on the browser for interpolating. The points, however, and bars are interpolated in JS.

The `interpolateRgb()` function is only used:
-- in `computeColor()` to get an interpolated color for a POINT (element of a LINE) between 2 stop points;
-- in `computeGradient()` to get an interpolated color for "max" & "min" points - to get "reference points" for a further calculation of a gradient (which is done by a browser).

The current situating (before this PR is merged) - there is a 3rd party "d3-interpolate" dependency.
This PR at least does not make it worse - it just adds some preliminary processing to convert `--var(...)` variables to real values.
So - very very IMHO - we may try merging this PR and then see if anything can be done better in a future.
I believe this is not a change which cannot be reverted/improved.

Issue #1248, opened 2025-05-14
State last fails to render card


MrWaggel:
Changing the state to `last` fails to render the card;

```
type: custom:mini-graph-card
group_by: date
hours_to_show: 168
aggregate_func: diff
icon: mdi:lightning-bolt
height: 80
entities:
  - entity: sensor.p1_meter_energy_import
name: Electricity Consumption
show:
  graph: bar
  state: last 
```

![Image](https://github.com/user-attachments/assets/86439c11-feab-4544-b520-b3022b717bc7)

```
Uncaught (in promise) TypeError: this.points[a] is undefined
    getEntityState mini-graph-card-bundle.js:1
    shouldUpdate mini-graph-card-bundle.js:1
    performUpdate mini-graph-card-bundle.js:1
    _enqueueUpdate mini-graph-card-bundle.js:1
    T mini-graph-card-bundle.js:1
    g mini-graph-card-bundle.js:1
    promise callback*T mini-graph-card-bundle.js:1
    f mini-graph-card-bundle.js:1
    U mini-graph-card-bundle.js:1
    U mini-graph-card-bundle.js:1
    _enqueueUpdate mini-graph-card-bundle.js:1
    requestUpdateInternal mini-graph-card-bundle.js:1
    set mini-graph-card-bundle.js:1
    set hass mini-graph-card-bundle.js:1
    value hui-card.ts:190
    performUpdate reactive-element.ts:1333
    scheduleUpdate reactive-element.ts:1262
    _$Ej reactive-element.ts:1237
    requestUpdate reactive-element.ts:1215
    set reactive-element.ts:731
    value hui-section.ts:132
    value hui-section.ts:131
    performUpdate reactive-element.ts:1333
    scheduleUpdate reactive-element.ts:1262
    _$Ej reactive-element.ts:1237
    requestUpdate reactive-element.ts:1215
    set reactive-element.ts:731
    value hui-view.ts:266
    value hui-view.ts:264
    performUpdate reactive-element.ts:1333
    scheduleUpdate reactive-element.ts:1262
    _$Ej reactive-element.ts:1237
    requestUpdate reactive-element.ts:1215
    set reactive-element.ts:731
    value hui-root.ts:556
    _$AE reactive-element.ts:1395
    performUpdate reactive-element.ts:1345
    scheduleUpdate reactive-element.ts:1262
    _$Ej reactive-element.ts:1237
    requestUpdate reactive-element.ts:1215
[...]

PR #1308, opened 2025-11-02
fix: state: last to use value from last Graph coordinate

whisdol:
Currently, `state: last` breaks rendering with bar charts (#1248) because `points` are not calculated for bar charts.
This fix accesses the value of the last coordinate from the `Graph` instead, which is exactly what I want to have displayed when using this option.
To my understanding (and testing), using the last `coord` is a simpler and more reliable fix than always calculating `points` (ref #1075), because the last `point` for a grouped bar chart is not the value of the last coordinate (the last `point` is actually the average of the last two states, so a completely different value). (Surprisingly, the same is true for grouped line graphs, where the point is selectable and shows this incorrect value as the hover state, but that may be a different issue.)

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -276,7 +276,7 @@ class MiniGraphCard extends LitElement {
   getEntityState(id) {
     const entityConfig = this.config.entities[id];
     if (this.config.show.state === 'last') {
-      return this.points[id][this.points[id].length - 1][V];
+      return this.Graph[id].coords[this.Graph[id].coords.length - 1][V];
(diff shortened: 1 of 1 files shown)


ildar170975:
@whisdol 
Please check another proposal - https://github.com/kalkih/mini-graph-card/pull/1322
Observations & critics are appreciated.

ildar170975:
Closed in favour of https://github.com/kalkih/mini-graph-card/pull/1322
Sincere thanks for contribution!

PR #1321, opened 2025-12-12
fix: style for icon[loc=right]

### LINKED ISSUE #1318: align_icon + show.name interaction
# Issue
When show.name is false, align_icon: right is not honored

### Config
ℹ️ entity chosen from 'standard' ones for the sample
```yaml
type: custom:mini-graph-card
entities:
 - sensor.sun_next_rising
show:
  name: false
align_icon: right
```
### Image
Image

### "As-is" expected
I'd expect it to display like this, at least, based on how other configuration combos work

Image

_Photoshopped to remove the name manually when `show.name: true`_

### _My_ ideal
Ideally, I'd like it to display like this, but understand if that's not feasible

Image

_Photoshopped to show personal preference of what I was _trying_ to get, when I found the bug of `align_icon` not being honored._

### Simulated with what is available
Image

_This alternative puts the value into the name, so avoided photoshop, but looks inconsistent next to other cards._

ildar170975:
When a name is hidden and `align_icon: right`, we can see an icon left-aligned:
```
    entities:
      - sensor.xxx
    show:
      name: false
    align_icon: right
```
image

After the fix:
image

### DIFF
--- a/src/style.js
+++ b/src/style.js
@@ -109,6 +109,9 @@ const style = css`
     margin-right: .6em;
     margin-left: 0;
   }
+  .icon[loc="right"] {
+    margin-left: auto;
+  }
   .icon[loc="state"] {
     align-self: center;
   }


akloeckner:
While this seems a straightforward fix, I'm wondering why `align_icon: state` also right-aligns but uses a different CSS. Should this be consistent? Or is there a reason, it's different?

ildar170975:
> While this seems a straightforward fix, I'm wondering why `align_icon: state` also right-aligns but uses a different CSS. Should this be consistent? Or is there a reason, it's different?

@akloeckner 
Sorry for a long delay.

What we have now:

align_icon: state
image

align_icon: left
image

align_icon: right
image

The `align_icon: state` case places an icon into the same container as a state.
Note that in `align_icon: left/right` cases the icon is placed into the same container as a name (which may be missing).
And if a state is missing with `align_icon: state` - then the icon is not shown at all:

image

I would say that this is normal that an icon is not shown - because we explicitly "bound" it to a state - which we decided not to show.

Also note that in RTL for `align_icon: state` an icon will be in a left, a state will be in a right side.

As for `align_icon: left/right` cases, they should be modified for RTL, this should be a different additional fix.

ildar170975:
ping @akloeckner

Issue #862, opened 2022-11-02
[Feature Request] Logarithm option on entity


smugleafdev:
Logarithmic is a setting for the card, but only the whole card. So if you have two entities, they are either both set to logarithmic, or neither are.

I am requesting this be a per entity setting either instead of, or in addition to, the full card setting.

rgnyldz:
I too need a similar thing. If you have two temperature sensors (inside around 23C and outside around 2C) you can have just two straight lines in the graph. So you can't visually compare the fluctuations. If you have some kind of logarithmic option for each entity you could see more than just two straight lines.

ildar170975:
> If you have two temperature sensors (inside around 23C and outside around 2C) you can have just two straight lines in the graph. So you can't visually compare the fluctuations

Why not making this simpler?
![Image](https://github.com/user-attachments/assets/a330dbc8-4040-4051-a0c1-af06e570ad7f)

rgnyldz:
You are one awesome dude :)

ildar170975:
check v0.14

PR #1348, opened 2026-03-09
feat: add per-entity logarithmic override

mxr:
Adds per-entity logarithmic configuration so mixed-scale graphs work correctly.

Specifying 'secondary' for the y-axis isn't enough for my use-case. I have a graph with two waveforms: one waveform is brightness on the primary y-axis (roughly 1 to 5000), while another waveform is sunrise/sunset on the secondary y-axis (boolean 0/1). With only global `logarithmic: true`, the sunrise/sunset waveform breaks; this change allows keeping brightness logarithmic while leaving sunrise/sunset linear via entity-level override.

Resolves https://github.com/kalkih/mini-graph-card/issues/862

| Current | Example of why `logarithmic: true` globally doesn't work | Example with the fix |
|--------|--------|--------|
|  |  |  |

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ The card works with entities from within the **sensor** & **binary_sensor** doma
 
 ## Install
 
-### HACS (recommended) 
+### HACS (recommended)
 
 This card is available in [HACS](https://hacs.xyz/) (Home Assistant Community Store).
 
@@ -210,7 +210,7 @@ color_thresholds:
   - value: 4
     color: "#0000ff"
 ```
-The example above will result in the following colors of the graph: if value is 
+The example above will result in the following colors of the graph: if value is
 * between `0` (including this value) and  `1.33333`, the color is `#ff0000`,
 * between `1.33333` (including this value) and `2.666667`, the color is `#ffff00`,
 * between `2.666667` (including this value) and `4`, the color is `#00ff00`,
@@ -258,6 +258,12 @@ These buckets are converted later to single point/bar on the graph. Aggregate fu
 | `sum` | v0.9.2 |
 | `delta` | v0.9.4 | Calculates difference between max and min value
 | `diff` | v0.11.0 | Calculates difference between first and last value
+
+### Logarithmic options
+
+Normally gaps between numbers on the graph are equal; the gap between 1 and 2 on the graph is the same as the gap between 100 and 101. The `logarithmic` option applies a [logarithmic transformation](https://en.wikipedia.org/wiki/Log_transformation_(statistics)) to the graph. With `logarithmic`, the graph is scaled by powers of 10, so the gap between 1, 10, 100, etc are equal. This is useful when your values span a wide range. Illuminance, for example, can swing from 1 to 5000 over the course of a day, and without a transformation it's hard to read the smaller values on the graph.
+
+Note that this option rounds up the input to 1 so negative numbers or numbers less than 1 are rendered as 0; this is different from the formal definition of logarithm, where `log(x) == 0` when `x<1` and $\infty$ when `x<0`.
 
 ### Theme variables
 The following theme variables can be set in your HA theme to customize the appearance of the card.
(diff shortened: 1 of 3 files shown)


ildar170975:
Thank you! Will check it soon.
Meanwhile, if you have some spare time, test other PRs; start with recently updated ones. Of course, only if you are interested in the proposed functionality, no obligations))

README.md
```diff
@@ -117,7 +117,7 @@ We recommend looking at the [Example usage section](#example-usage) to understan
 | smoothing | boolean | `true` | v0.8.0 | Whether to make graph line smooth.
 | state_map | [state map object](#state-map-object) |  | v0.8.0 | List of entity states to convert (order matters as position becomes a value on the graph).
 | value_factor | number | 0 | v0.9.4 | Scale value by order of magnitude (e.g. convert Watts to kilo Watts), use negative value to scale down.
-| logarithmic | boolean | `false` | v0.10.0 | Use a Logarithmic scale for the graph
+| logarithmic | boolean | `false` | v0.10.0 | Use a logarithmic scale for the graph (default for all entities, can be overridden per entity).
```
ildar170975: Please remove "(...)", we do not have similar remarks for other options.

src/main.js
```diff
@@ -646,6 +646,14 @@ class MiniGraphCard extends LitElement {
     return this.config.entities.filter(entity => entity.show_graph !== false);
   }
 
+  entityUsesLogarithmic(index) {
```
ildar170975: Please:
1. Rename it to smth like `computeUsesLogarithmic` (or find a better wording) - similarly to `computeColor()`, `computeUom()` etc.
2. Place it before `computeColor()`, let's have these `get xxx` getters be together.

ildar170975: Also, please add this description:
```
  /**
  * Checks if the entity uses a logarithmic scale
  * @param {number} index Index of an entity in config.entities
  * @returns {boolean} True - the entity uses a logarithmic scale, false - otherwise
  */
```
or with a better wording.

ildar170975:
@mxr 
Added some observations, please consider.

For me, the whole "graph" functionality (described in a `graph.js`) is still a complex thing which I am not 100% familiar with.
In this PR there are mostly SIMPLE changes, a global `logarithmic` option is "augmented" by a a per-entity option.
What probably needs to be at least confirmed - is a building a logarithmic graph itself.
Of course, we may blindly trust the current implementation - but these algorithms should be analysed & improved if needed in future (at least to become "maintainable").

For now, I think this PR achieved it's minimal goal (per-entity option).
@akloeckner - please review when you have a spare time.

mxr:
Thanks for the feedback. I implemented it in https://github.com/mxr/mini-graph-card/commit/6518e8b5d73e4a4bb44140c9885ae3091130fc25 (GitHub seems to be lagging so I amended the prior commit)

> What probably needs to be at least confirmed - is a building a logarithmic graph itself.

I have an example in the PR description. Let me know if I misunderstood.

ildar170975:
I totally understood your example, I meant a technical implementation of the Graph class itself. Repeat, the PR looks fine, but since personally me may not know some details of Graph class, I am not sure that the whole logarithmic thing works properly. Let’s merge and then test.

ildar170975:
ping @akloeckner

ildar170975 (approved):

ildar170975:
@mxr 
I approved these changes on March, just did not push the "approve" button )))
We have a bunch of unmerged PRs now, let's wait for @akloeckner (we do things in a team here) & make a RELEASE.
Thanks again for the contribution, sorry for the delay.
As you probably know, you can start using js with your changes right now (download it from here - https://github.com/kalkih/mini-graph-card/actions/runs/22955326308/job/66630389391)

ildar170975:
@mxr 
Based on personal experience, I think we need to add some clarifications to Docs about "what a logarithmic thing is".
(BTW, for stock History graph & Statistics graph cards there were even plans to exclude a corresponding "logarithmic scale" option from UI editor & make it available for yaml only - many people checked this option & did not understand what they got).
At least, we need to add a note about "for any values < 1 - the logarithmic graph will be 0" - this is a current implementation, i.e. here yellow values fall back to 0:
изображение

mxr:
@ildar170975 thanks added some docs

ildar170975:
> @ildar170975 thanks added some docs

Thanks, I think these clarifications are good, they seem to be clear even for non-technical people.

PR #1233, opened 2025-04-23
feat: Add more options to customize a number format, use default accuracy & format settings

ildar170975:
1. Define `decimals` options per entity.
2. Override a card-wide `decimal` option for each Y-axis by using new `decimals_primary_labels` & `decimals_secondary_labels`.
3. If no `decimals` options are defined (card-wide or per entity) - use default entity accuracy settings (individual entity accuracy settings are accounted).
4. Use HA frontend number format settings.

PR #1350, opened 2026-03-10
feat: Use native localized units & order "value - delimiter - unit"

### LINKED ISSUE #1287: [FR] Hide unit for Min/Avg/Max value
It would be nice if it was possible to hide the unit but only for the `show: average` and `show: extrema`  value.

For example, I have a graph for the temperature outside that I would like to show the unit (C) for the main value but hide the unit for the min/avg/max value under it (which would be the same unit anyway)

It would help legibility on small screen such as mobile, especially when the unit is `hPa` or `UV Index` (something larger).

Thanks!

ildar170975:
Adds more features to https://github.com/kalkih/mini-graph-card/pull/1233.
So, suggest to test along with https://github.com/kalkih/mini-graph-card/pull/1233 merged.

Additionally - implements https://github.com/kalkih/mini-graph-card/issues/1287.

---

**Features:**

Support of native units for attributes; example - temperatures in `climate.ecobee` of Demo integration:

image

---

Support of native presentation of "percentage" values: some locales like “German” demand a presence of a whitespace between a value & "%", some like “English” do not place a whitespace:

image

---

Support of native localized units dependently on a locale.
Support of native reversed order for a state/attribute value presentation dependently on a locale.
Example - monetary values:

Language: "English (GB)"
Number format: "Use system locale"

image

Language: "English"
Number format: "Use language settings"

image

---

Support of translated units; example - Github integration:
Language: “German”

image

---

Allow to hide a unit for extrema & average info for a compactness (https://github.com/kalkih/mini-graph-card/issues/1287):

image

### DIFF
--- a/src/locale.js
+++ b/src/locale.js
@@ -20,7 +20,6 @@ const blankBeforePercent = (
   }
 };
 
-// eslint-disable-next-line import/prefer-default-export
 export {
   blankBeforePercent,
 };
--- a/src/locale.js
+++ b/src/locale.js
@@ -1,7 +1,7 @@
 /**
  * Checks if a whitespace is needed before a "%" unit dependently on a locale
  * @param {FrontendLocaleData} localeOptions Object containing
- * a user-selected language and formatting
+ * a user-selected language and formatting settings
  * @returns {string} Whitespace if needed before "%", empty otherwise
  */
 const blankBeforePercent = (localeOptions) => {
--- a/src/locale.js
+++ b/src/locale.js
@@ -8,15 +8,15 @@ const blankBeforePercent = (
   localeOptions, // FrontendLocaleData
 ) => {
   switch (localeOptions.language) {
-    case "cs":
-    case "de":
-    case "fi":
-    case "fr":
-    case "sk":
-    case "sv":
-      return " ";
+    case 'cs':
+    case 'de':
+    case 'fi':
+    case 'fr':
+    case 'sk':
+    case 'sv':
+      return ' ';
     default:
-      return "";
+      return '';
   }
 };
 
--- a/src/locale.js
+++ b/src/locale.js
@@ -20,6 +20,7 @@ const blankBeforePercent = (
   }
 };
 
+// eslint-disable-next-line import/prefer-default-export
 export {
   blankBeforePercent,
 };
--- a/src/locale.js
+++ b/src/locale.js
@@ -1,12 +1,10 @@
-// This function is taken from HA Frontend
 /**
  * Checks if a whitespace is needed before a "%" unit dependently on a locale
- * @param localeOptions Object containing a user-selected language and formatting
+ * @param {FrontendLocaleData} localeOptions Object containing
+ * a user-selected language and formatting
  * @returns {string} Whitespace if needed before "%", empty otherwise
  */
-const blankBeforePercent = (
-  localeOptions, // FrontendLocaleData
-) => {
+const blankBeforePercent = localeOptions => {
   switch (localeOptions.language) {
     case 'cs':
     case 'de':
--- a/src/locale.js
+++ b/src/locale.js
@@ -4,7 +4,7 @@
  * a user-selected language and formatting
  * @returns {string} Whitespace if needed before "%", empty otherwise
  */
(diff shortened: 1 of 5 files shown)


ildar170975:
A possible improvement which can/should be done in future:
perhaps we need to define an order "value - delimiter - unit" not on every render - but once we get a meaningful state (i.e. not `unknown` or `unavailable`), may slightly increase a performance.

src/locale.js
```diff
@@ -0,0 +1,23 @@
...
+ * @returns {string} Whitespace if needed before "%", empty otherwise
+ */
+const blankBeforePercent = (localeOptions) => {
+  switch (localeOptions.language) {
+    case 'cs':
+    case 'de':
+    case 'fi':
+    case 'fr':
+    case 'sk':
+    case 'sv':
```
jlsjonas, line 572: Could we potentially extract this from home assistant's api instead?
If not, are you sure this list is exhaustive? Would be good to refer to the (preferably HA) source to maintain over time.

src/main.js
```diff
@@ -769,6 +790,79 @@ class MiniGraphCard extends LitElement {
...
+  * Returns settings defining an order of a state/attrubute value presentation
+  * @returns {Object}
+  * directOrder - true: "value literal unit", false: "unit literal value";
+  *
+  * delimiter - an optional literal separator between value & unit
+  * @param index Index of an entity in config.entities
+  */
+  computeStateOrder(index) {
+    const entityId = this.config.entities[index].entity;
+    const { attribute } = this.config.entities[index];
```
jlsjonas, line 1014:
```suggestion
    const { entityId, attribute } = this.config.entities[index];
```

src/main.js
```diff
@@ -708,19 +707,41 @@ class MiniGraphCard extends LitElement {
...
-        : (
-          this.config.unit !== undefined
-            ? this.config.unit
-            : (
-              !this.config.entities[index].attribute
-                ? (this.entity[index].attributes.unit_of_measurement || '')
-                : ''
-            )
-        )
-    );
```
jlsjonas, line 866: Note that this format was originally used for performance reasons (we also run on lower-spec devices), please check the performance impact of the change(s)

jlsjonas (commented):
Not a thorough review yet, thanks for your continued contributions & glad to see your progress!

src/locale.js
ildar170975: > Could we potentially extract this from home assistant's api instead?

I do not think there is such a "HA api" (((. We can use methods of `hass` object (specially exposed for external use) - but there are no exposed methods for formatting.
Recently I added methods to `hass` to represent parts of a state & attribute - earlier only WHOLE strings for a state & attribute were available (i.e. "value with unit", not separately). Now I am thinking to expose `formatNumber()` (internal frontend function) to the outer world. But meanwhile we have to manually "copy" functions from Frontend to get a similar result.

> If not, are you sure this list is exhaustive? Would be good to refer to the (preferably HA) source to maintain over time.

No guarantee that this list is full. The best scenario would be using an exposed HA function - which is not available now, see above.

src/main.js
ildar170975: The problem is that `entities[index]` does not have the `entityId` option.
Perhaps we can do smth like
```
const { entity: entityId, attribute } = this.config.entities[index];
```
just a speculation, since I know a little about JS.

My approach was simple: when I get a warning about a "deconstruction", I use that "`const { xxx } = yyy`" syntax, otherwise - use "old classic" way.

src/main.js
ildar170975: Yeah, using functions like `formatEntityAttributeValueToParts` needs more time.
Same as for "use `states['sensor.xxx'].state`" vs "use `formatEntityStateToParts()`".
But we need to follow HA frontend number formatting settings, individual entity precision settings, translations...
My only hope is that `computeState()` (same for `computeUom()`) is only used for presentation, i.e. not being called very often.

ildar170975:
@jlsjonas 
Hi Jonas, nice to see you again). Thanks for the review, please go on)))

ildar170975:
Additional fix - https://github.com/kalkih/mini-graph-card/pull/1369

PR #1364, opened 2026-06-02
feat: 2026.6 card suggestion

ildar170975:
https://developers.home-assistant.io/blog/2026/05/27/custom-card-suggestions

image

PR #1365, opened 2026-06-03
feat: Add card picker suggestion for numeric sensors

piitaya:
Mini Graph Card now suggests itself in the card picker when you pick a
numeric sensor (temperature, humidity, power, energy, and so on). It shows
up under the Community section, next to the built-in suggestions.

Requires Home Assistant 2026.6 or later.

Docs: https://developers.home-assistant.io/docs/frontend/custom-ui/custom-card#suggesting-your-card-for-an-entity

CleanShot 2026-06-03 at 11 17 03

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -1079,6 +1079,18 @@ class MiniGraphCard extends LitElement {
 
 customElements.define('mini-graph-card', MiniGraphCard);
 
+const NUMERIC_DOMAINS = ['counter', 'input_number', 'number'];
+
+const isNumericEntity = (hass, entityId) => {
+  const domain = entityId.split('.')[0];
+  if (NUMERIC_DOMAINS.includes(domain)) return true;
+  if (domain !== 'sensor') return false;
+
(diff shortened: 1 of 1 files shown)


ildar170975:
@piitaya 
Hi))
Thanks for the contribution. 
We already have the unmerged https://github.com/kalkih/mini-graph-card/pull/1364
Can you take a look at it, it has some more numerical domains added; if you add them, we will merge your PR instead.

piitaya:
I added support for more domain. However, I don't think zone is a good idea. Technically it's a number but graph is not a good card to represent a zone. A map card would be more appropriate.

Also, HA frontend doesn't consider zone as numerical entities in general.

ildar170975:
> Technically it's a number but graph is not a good card to represent a zone

Agreed. 
Closed another PR in favour of this one. 
Gonna test and approve then; will be merged along with other 1st priority PRs soon.

ildar170975 (approved):

Issue #153, opened 2019-09-13
feature request: per entity thresholds


wrboyce:
As a concrete example, I have a card showing my Download/Upload/Ping: it would be nice for each line to have its own appropriate colour thresholds.

hopperjef:
Was just searching for this very thing, hope this feature gets added soon

Psicrow925:
searching for this ability too, could you please add this feature :)

qupterra:
I'd also be interested in having this feature available. 👍

Nickduino:
> I have a card showing my Download/Upload/Ping: it would be nice for each line to have its own appropriate colour thresholds.

That's exactly my use case and I came to the same conclusion.

spydoor17:
I would also be grateful if this feature was provided.  
My use case is for a water heater where I'd like to see:
Water temp (lower 30, upper 70) (blue to amber to red)
Energy consumed (lower 0, upper 6000) (amber to red)

Matt2uk:
I'd like the feature too.  I've a graph for my Air Quality Meter and it would be good to be able to colour code sensor in accordance with it's Green/Amber/Red state.

CalicodesIles:
Also interested by this feature request ! Thank you !

CalamarBicefalo:
Also was hoping to do this. In case it helps I'll share my use case:

I intend to overlay all Air Quality metrics on 1 graph, and each has different thresholds (PM2.5 PM10 VOC...)

Ultimately I care more about _something_ being red or green, whilst it is nice to keep the focus on hover in case I want to dig out a specific metric.

ildar170975:
> PM2.5 PM10 VOC...

Don’t these parameters have different ranges? Perhaps they should not be placed on the same axis, maximum two axis are available

SgtFlippy:
I'm not sure since I'm finding contrary information, but anyways, I'm not getting it working, since this is still open, I guess it's still a request?

I'll add my use case: I'm trying to make a graph that shows neatly when we're using too much power within 15 minutes (peak tariff).
So I have a line of the current peak that's being generated, a line that shows the highest peak we made this month (since for the rest of the month, that's the highest we should go) and a line with the base-line 2.5 kW that's included in the contract.

I would like to have the current generated peak, to change colour when it's going near 2.5. Ideally I could template it so that it changes colour when nearing the current month's highest peak, but that might be dreaming.

ildar170975:
[...]

PR #1383, opened 2026-06-27
feat: color_thresholds per entity

ildar170975:
Add per-entity `color_thresholds`.
(imho not a rather critical feature, it may make a picture less readable, more confusing)

Solves https://github.com/kalkih/mini-graph-card/issues/153


HirschBerge:
Couldn't we get a pre-release with this?

ildar170975:
> Couldn't we get a pre-release with this?

soon, please wait, a couple of weeks

PR #1399, opened 2026-07-11
feat: Use line & bar graphs together

ildar170975:
Allow to use `line` & `bar` graphs together:

image

---

Done:
1. Added a support of combined graphs.
2. Introduced a new `graph` option (`line`, `bar`) to define an individual graph type:
```
...
entities:
  - entity: sensor.xxx
    graph: bar
...
```
3. Changed a logic to define margins for a graph container.
4. Changed mouse events processing to allow to select bars & points.
5. A label is not shown for a "static_value" graph if this graph is "bars".

----

Changes in Docs:
1. (unrelated) "stacked bars" renamed to "overlapping bars".
2. Added a new "graph" per-entity option.
3. Added a new "Graph types" chapter.
4. Added examples for different graph types.
5. Added a remark about showing a labels for static value's bars.
6. Naming: "static line" is renamed (where it is needed) to "static value" since a static value can be presented as bars.

---

Additional unrelated changes:
1. Move a check for `MAX_BARS` from `buildConfig()` to `setConfig()` (to utilize card class methods).
2. Improve processing of cached properties (check for `undefined` instead of `null` for consistency).
3. Add an initialization for `_graphMargin` (not critical, just for consistency).
4. Add comments in `getEntityState()`.


fran68:
Hi ildar,
I've also played around in the past a little bit with that topic and made a simple approach. If you configure e.g. `hours_to_show = 8` and `points_per_hour = 1 ` you will get 8 data points. That means 8 segments in case a bar is drawn. However for lines you only get seven segments (with 8 points since the first and last data points lies on the edge). So I decided to read an additional data point from history (9 datapoints). Then, when a bar is shown the first data point is stripped. This results in 8 segments/bars. A line is drawn with all 9 datapoints covering 8 segments. So line and bar are in sync. There are some minor adaptions for the tooltip and to get the proper min and max values.  

My plan is to align/move the data points of the line in dependence on the chosen aggregate_func. E.g., an average value (aggregate_func avg) should be in the middle of the segment and not at the end which is now the default for all aggregate functions.

Attached a figure how it looks like in my local copy. The x-axis make it transparent.
image

If you like I can share. Maybe you've already found a better approach.

Another example with BezierC which has its advantages regarding the values.
image

ildar170975:
@fran68 
Hi! I am away from PC, do you have an account in HA community? If yes - can you contact me via DM or a chat, I am ildar_Gabdullin there.

ildar170975:
@fran68 
> If you configure e.g. hours_to_show = 8 and points_per_hour = 1 you will get 8 data points. That means 8 segments in case a bar is drawn. However for lines you only get seven segments (with 8 points since the first and last data points lies on the edge). So I decided to read an additional data point from history (9 datapoints). Then, when a bar is shown the first data point is stripped. This results in 8 segments/bars. A line is drawn with all 9 datapoints covering 8 segments. So line and bar are in sync

This was changed in https://github.com/kalkih/mini-graph-card/pull/1439 which I earlier mentioned in our chat.
Now we have (as an example) 9 points for a line graph (incl. the initial value on a left boundary) & 8 bars for a bars graph.

> There are some minor adaptions for the tooltip and to get the proper min and max values.

This was changed in https://github.com/kalkih/mini-graph-card/pull/1447 - a `coords` property is reduced for bars right in `update()` method to get proper max/min values.

> My plan is to align/move the data points of the line in dependence on the chosen aggregate_func. E.g., an average value (aggregate_func avg) should be in the middle of the segment and not at the end which is now the default for all aggregate functions.

Not sure a "line point" should be in a middle of a "bar".
Check this:

WARNING: the text below was edited because I found a bug in `setTooltip()` which caused issues & then fixed it in https://github.com/kalkih/mini-graph-card/pull/1449.

image

```
type: custom:mini-graph-card
entities:
  - entity: sensor.system_monitor_processor_use
  - entity: sensor.system_monitor_memory_use
    show_state: true
    y_axis: secondary
    graph: bar
height: 200
y_axis:
  primary:
    lower_bound: 0
smoothing: false
bar_spacing: 0
points_per_hour: 3
hours_to_show: 2
show:
  labels: true
  labels_secondary: true
  points: true
  fill: true
  name: false
  icon: false
```
Here a "line point" is aligned with a RIGHT edge of a "bar" - with `bar_spacing: 0`.
"Aligned" = MATHEMATICALLY aligned when `bar_spacing: 0` & there is only one bar graph, see formulae in `calcPoints()` for calculating coord.x & in `getBars()` for calculating bar.x.

In case of a bigger `bar_spacing` we can get this misalignment:

image

~And even with `bar_spacing: 0` and more than 1 bar graphs we'll get a misalignment as well.~ Sorry, that was a hasty thought, a point will be aligned with a right edge of the LAST bar in a group.
Imho this is not an end of days, a user is fully responsible for an inputted config - just "do not hold it this way", i.e. do not set a huge `bar_spacing`.

Anyway, imagine we have a hard-coded `bar_spacing: 0` & always only 1 "bars" graph.
Then a "line point" is aligned with a RIGHT edge of a "bar" as was shown above.
I would say this is CORRECT: a bar or a point are calculated based on `coords[i]` which represents some bucked - which is an aggregated value (max, min, last, average, ...) - and I would say that a RIGHT edge is a proper presentation, it means "end of a bucket's interval", and visually a bar may be considered as a "bucket's interval".

PR #1409, opened 2026-07-24
feat: add support of zeroline/baseline, bars with negative value

fran68:
Hi, this PR introduces a zero line (called baseline in code) which allows bars with negative values. The baseline is also the reference for the fill which is drawn in a line graph and for the gradient of the fill which now fades out to this baseline which is the zero line by default. To be honest the logarithmic option needs some testing.

Unfortunately, it overlaps a little bit with the merge of PR kalkih#1205. But with a small change you can also achieve the customization for the fill. 

Cached getters are used. These are only updated when there is a change for the bounds (min/max setters called by main.js). That affects also the _calcY function which is called multiple times.
Not sure about what is the best way to get values from Graph[i] from outside, by getter or defining a function which keeps all the data which are specific to the axis (max, min, baseline, yRatio, bRatio)

I tried to maintain the look of the bars.

- only zero values: (zero) line stays at the bottom. Margin (zero margin) at the bottom and the top
- only positive values, or positive and zero values. Margins at the top and bottom (zero margin).
- only negative values, or negative and zero values. Margins at the top (zero margin) and bottom.
- sign values (positive and negative values): Margins at the top and bottom. If there is a zero value no bar is drawn at the zeroline/baseline.

image
image
image

### DIFF
--- a/src/graph.js
+++ b/src/graph.js
@@ -41,6 +41,11 @@ export default class Graph {
     this.margin = margin;
     this._max = 0;
     this._min = 0;
+    this._lmax = null;
+    this._lmin = null;
+    this._yRatio = null;
+    this._baseline = null;
+    this._bRatio = null;
     this.points = points; // stands for "points_per_hour"
     this.hours = hours; // stands for "hours_to_show"
     this.aggregateFuncName = aggregateFuncName;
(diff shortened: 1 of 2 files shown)


fran68:
I've updated the var names.

ildar170975:
@fran68 
Closing in favour of https://github.com/kalkih/mini-graph-card/pull/1443
Mostly the same method is used in https://github.com/kalkih/mini-graph-card/pull/1443 (please check the 1st post in that PR).
Of course some credits are yours!

Basic differences are (incl. not mentioned in that PR):
-- no caching for min/max/baseline/baselineRatio values;
-- added support for a future `invert` option;
-- a `width` field is not repeated for every bar in a corresponding data array.

Issue #641, opened 2021-08-11
Longterm Statistics Support


alexpomsft:
This is a request for the mini-graph-card support the min/max/mean in addition to the "current" values for statistics enabled sensors. See the [statistics graph card](https://www.home-assistant.io/lovelace/statistics-graph/) for examples. I find the mini-graph-card to be super nice and it would be great if we could add min/max/mean behind/under the current values for these sensors.

jlsjonas:
Perhaps someday, would require a significant rewrite in how entities are handled. (allowing multiple graphs per _entity_, statistics flag to switch to [Different API](https://github.com/home-assistant/frontend/blob/ff2fa9a78cbe5788e71aba2e7c60fc952935b1c4/src/data/history.ts#L355) for fetching the required data; reworking cache & pretty much all points touching this data)

PR #1423, opened 2026-08-09
feat: read a series from statistics

Olen:
## Summary

Adds a `statistics` option, per entity or card-wide, that sources a series from Home Assistant's long-term statistics instead of raw recorder history.

```yaml
type: custom:mini-graph-card
hours_to_show: 336
points_per_hour: 1
entities:
  - entity: sensor.outside_temperature
    statistics: true          # or: { period: hour, type: mean }
```

## Rationale

The card has one data path: `history/period`, i.e. raw recorder history. That has two consequences for long graphs.

**It is bounded by the recorder.** With `purge_keep_days: 30`, a graph cannot show more than 30 days no matter what `hours_to_show` says. Long-term statistics are kept indefinitely, so ranges the recorder simply cannot serve become possible — on my install these sensors have hourly statistics back to 2022 against 30 days of raw history.

**It scales with state rows, not with pixels.** Every state row in the window is fetched and then bucketed client side, so cost is driven by how chattily the sensor reports rather than by how much the graph can actually display. A 336 h graph on a ~300 px card can show a few hundred points; fetching six figures of rows to draw them is pure overhead. Statistics are pre-aggregated server side, so the fetch is already close to what gets drawn.

This matters most exactly where the card is least usable today, and it is opt-in — nothing changes unless `statistics` is set.

## Example: a real dashboard

A plant dashboard with 10 cards, each a 4-tab set (1 h / 6 h / 24 h / 14 d) over 2 entities — 40 card instances, 80 series, all fetched when the view opens.

Measured per tab, via `history/history_during_period` for all 10 cards:

| tab | recorder time | rows returned |
|---|---|---|
| 1 h | 0.10 s | 473 |
| 6 h | 0.16 s | 2,749 |
| 24 h | 0.49 s | 11,815 |
| **14 d** | **13.26 s** | **152,087** |
| **total** | **14.01 s** | **167,124** |

The 14-day tab is **95 % of the page's history cost**. The same 20 series over the same 14 days, via `recorder/statistics_during_period` at `period: hour`:

| source | time | points |
|---|---|---|
| raw history | 13.26 s | 152,087 rows |
| hourly statistics | **1.04 s** | 6,700 points |

**~13x faster for the same window**, and visually identical: 336 hourly points per series is already more than a 300 px-wide graph can resolve.

Because the option is per entity/card, the short tabs can stay on raw history where sub-hour resolution actually matters, and only the long tab switches:

```yaml
# 1h / 6h / 24h tabs — unchanged, raw history
# 14d tab:
hours_to_show: 336
points_per_hour: 1
statistics:
  type: mean
```

For reference, a full year of the same 20 series at `period: day` costs 3.0 s / 3,864 points — a range raw history cannot produce at all under a 30-day purge.

## Implementation

The fetch is the only thing that changes. Both paths produce the same normalised `[{last_changed, state}]` series, so bucketing, extrema, colour thresholds, bounds and the secondary y-axis are untouched. The shared tail of `updateEntity()` is extracted into `applyHistory()` and called from both.

- `period` defaults from `hours_to_show`, since HA only retains `5minute` statistics for a limited window (10 days by default) while hourly are kept indefinitely — so a long graph does not ask for data that has been purged.
- `type` selects `mean` (default), `min`, `max`, `sum` or `state`.
- Validation lives in `checkOption.js` as `checkStatistics()`, alongside the existing option checks.
- Statistics refetch the whole window rather than appending to the cached tail: a period's bucket is revised until that period completes, so an incremental merge would duplicate or truncate the boundary bucket. Cache separation is automatic, as the key already includes the config hash.

Entities without a `state_class` have no statistics; that is documented rather than silently falling back, so the failure is visible.

Refs #641

## Notes

- No new lint errors. `dev` currently reports 3 (`checkOption.js` x2 `no-param-reassign`, `others.js` missing semicolon); this branch reports the same 3.
- README documents the option, including the interaction with `points_per_hour` (statistics are aggregated into fixed periods, so more than one point per period only yields empty buckets).

### DIFF
--- a/src/buildConfig.js
+++ b/src/buildConfig.js
@@ -18,6 +18,7 @@ import {
   checkIntegerOption,
   checkBounds,
   checkColorThresholds,
+  checkStatistics,
 } from './checkOption';
 import { getFactor } from './others';
 
@@ -221,6 +222,9 @@ export default (config) => {
         );
       }
     }
+    // `statistics` may be set per entity or card-wide, and accepts a bare
+    // `true` as shorthand for the defaults.
+    checkStatistics(conf, i);
   });
   /* eslint-enable no-param-reassign */
 
--- a/README.md
+++ b/README.md
@@ -335,7 +335,7 @@ entities:
 | Option | Type | Default | Description
 |--------|------|---------|------------
 | period | string | derived from `hours_to_show` | `5minute`, `hour`, `day`, `week`, `month` or `year`.
-| type | string | `mean` | Which statistic to plot: `mean`, `min`, `max`, `sum` or `state`.
+| type | string | see below | Which statistic to plot: `mean`, `min`, `max`, `sum`, `state` or `change`.
 
 ```yaml
 entities:
@@ -351,8 +351,17 @@ the defaults, and `statistics: false` on an entity opts it back out.
 
 Notes:
 
-- Only entities with a `state_class` have statistics. Others produce an empty
-  graph, so leave them on raw history.
+- Only entities with a `state_class` have statistics; the card logs a warning
+  if there are none for an entity.
+- Which types exist depends on the `state_class`: `measurement` has `mean`,
+  `min` and `max`, `measurement_angle` has `mean` only (a circular mean),
+  `total` and `total_increasing` have `sum`, `state` and `change`. The card
+  reads the available types from a response, so a type which is not there is
+  replaced by a default one, with a warning in a console.
+- When `type` is not set, `mean` is used if it is available, otherwise `state`
+  - the reading a raw history would show. Use `change` to plot a consumption
+  per period instead, which is usually what is wanted for an energy, water or
+  gas sensor.
 - Home Assistant retains `5minute` statistics for a limited window (10 days by
   default) and hourly statistics indefinitely. When `period` is not set it is
   derived from `hours_to_show` so a long graph does not ask for data that has
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ We recommend looking at the [Example usage section](#example-usage) to understan
 | group_by | string | `interval` | v0.8.0 | Specify type of grouping of data, dynamic `interval`, `date` or `hour`.
 | update_interval | number |  | v0.4.0 | Specify a custom update interval of the history data (in seconds), instead of on every state change.
 | cache | boolean | `true` | v0.9.0 | Enable/disable local caching of history data.
-| statistics | boolean *or* object |  |  | Read the series from long-term statistics instead of raw history, see [Long-term statistics](#long-term-statistics).
+| statistics | boolean *or* object |  |  | Read the series from statistics instead of raw history, see [Statistics](#statistics).
 | show | list |  | v0.2.0 | List of UI elements to display/hide, for available items see [available show options](#available-show-options).
 | animate | boolean | `false` | v0.2.0 | Add a reveal animation to the graph.
 | height | number | `150` | v0.0.1 | Set a custom height of the line graph.
@@ -159,7 +159,7 @@ properties of the Entity object detailed in the following table (as per `sensor.
 | show_static_inactive | boolean |         | Set to true to disable hiding the line when a point of a line of another entity selected; meaningful for a [static line](#static-lines) only.
 | state_adaptive_color | boolean |         | Make the color of the state adapt to the entity/static value color.
 | y_axis | string |         | If 'secondary', displays using the secondary Y-axis on the right.
-| statistics | boolean *or* object |         | Override long-term statistics for this entity only, see [Long-term statistics](#long-term-statistics).
+| statistics | boolean *or* object |         | Override statistics for this entity only, see [Statistics](#statistics).
 | fixed_value | boolean |         | Set to true to graph the entity's current state as a fixed value instead of graphing its state history.
 | smoothing | boolean |         | Override for a flag indicating whether to make graph line smooth.
 | logarithmic | boolean |         | Override logarithmic scaling for this entity only (see [Logarithmic options](#logarithmic-options)).
@@ -313,15 +313,15 @@ These buckets are converted later to single point/bar on the graph. Aggregate fu
 | `delta` | v0.9.4 | Calculates difference between max and min value
 | `diff` | v0.11.0 | Calculates difference between first and last value
 
-### Long-term statistics
+### Statistics
 
 By default a series is read from the recorder's raw history. That history is
 bounded by the recorder's `purge_keep_days`, and it gets expensive over long
 spans - every state row in the window is fetched and then bucketed client side.
 
-Setting `statistics` reads the series from Home Assistant's long-term
-statistics instead. Those are pre-aggregated server side and kept far longer
-than raw history, so a wide graph costs a fraction of the rows.
+Setting `statistics` reads the series from Home Assistant's statistics
+instead. Those are pre-aggregated server side and kept far longer than raw
+history, so a wide graph costs a fraction of the rows.
 
 ```yaml
 type: custom:mini-graph-card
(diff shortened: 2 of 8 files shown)


ildar170975:
Sorry, did you use AI here?

Olen:
Yes, I use Claude Code to do the coding, while I herd it and ensure it behaves as I want.
It should be mentioned that I am more experienced in Python than in Javascript, so it could be that some issues have slipped past my review.  Just let me know, and I will look into that.

However, I have tested this, and can verify that it made a significant boost in loading a few graph-heavy dahsboards I use (showing graphs up to 14 days back in time), where they went from barely usable to quite snappy.

ildar170975:
I myself do not appreciate using AI for coding.
We cannot verify this PR only by “output result = expected result”, it is needed to verify the code. 
So, it will take time.

Olen:
That is ok.
It should hopefully be quite clean and easy to follow.

It just use the long term statistics to pull the data instead of the short term recorder _if_ you add the new config option.  It changes no parsing or handling of the actual data, just the source for the data.

ildar170975:
Did you actually check a correctness of a system call retrieving data from a server, and then generating a history?
“Check” - means a check of a code first, then a data themselves using debug tools.
Since you are a developer, you know how to debug a system. 

Converting to a draft since I see no proof yet. See, we cannot just add some AI-generated something only to set a mark “we support stats”.

Olen:
I have checked the data from the WS-endpoint (real documentation of the API is thin, but verfication with both the source code and with real data has been done).  Since there are no unit-tests in the repo, I can not systematically verify that this does not change anything, but the code only touches a few lines of current functions, and is mostly just adding the new data source, so I think the manual review should be good enough.

Since the long term statistics are calculated by HA using a time-weightet mean, the graphs from data in the recorder and a graph for the exact same period using long term statistics will not match 100%, but that is expected. 
A small challenge could be if one tries to use the long term statistics for short timespans (e.g. 1 hour).  Because the statistics are generated at the end of each period, they will not contain the newest values, leaving the graphs with a "gap" in the end.  This does not matter much for long term graphs, as missing the data from the last _hour_ in a _14 day_ graph is probably not a big deal.  But missing the last 5 minutes from a 30 minute graph is probably not what you want.  This is also why the `statistics` option should definitely not be the default.  It is genuienly useful for showing long time series and old data.  Not for real time tracking.

I have also verified that the `state` element in the card pulls data directly from the sensor in normal cases.  This can lead to inconsistency, where the last point on the graph shows one value, while the state is something else, or one could use `show.state: last` in the config, which will actually use the last value from the statistics, but when using the long term statistics, this value could be over one hour old.  You have to pick your poison here. 

Also - HA supports the option of changing units for a sensor. And integrations sometimes changes them as well.  When that happens, a "repair" is raised in HA to convert the old data in the long term statistics as well, but the user does not NEED to do it.  In that case, it could be a mismatch between the values in the long term statistics and the new values.  I consider this as a user-problem, and not really something that can be handled by the code here.  It should also not cause any real problems, other than graphs showing data based on old units.  

There are a few other caveats regarding the code itself: 
The long term statistics report the values as _float_ while the recorder returns _strings_.  All consumers of the data converts them to `Number` and/or use `parseFloat` (which makes sense, since we are generating graphs from them), so it should not cause any issues.  I am a bit hesitant to cast the returned floats to strings, just to convert them back to Number again a few lines later in the code just to make the returned values the same type.

Timestamps from the recorder include microseconds.  The long term statistics skip this, so they only have one second resolution.  Again, the consumers use `new Date()` which handles both formats fine.

And talking about timestamps.  I currently use bucket start as the timestamp for the value.  It could be argued that the midpoint in the bucket is a better choice.  Just a visualisation inconsistency, but I am open for suggestions.

A few things I think I will add:
- Long term statistics can not work with `state_map`.  Those maps only make sense for binary sensors or other sensors with non-numeric values.  Those are not stored in the long term statistics, so I should probably log an error message if someone tries to combine `statistics` and `state_map` in an entity.
- The same with attributes.  They are not stored in the long term statistics, and it is probably better to raise an error than to silently drop it.

ildar170975:
Away from PC, not easy to read your message, just a small remark:

> Since there are no unit-tests in the repo

There are test files in a corresponding folder; tests are disabled since there are lot of them (for datetime particularly), running them on every build will take time. The tests can be used in a local environment. 
As for testing of stats, the only possible way I see is using a pre-generated DB file containing required tables.

Olen:
>  Since there are no unit-tests in the repo, I can not systematically verify that this does not change anything, but the code only touches a few lines of current functions, and is mostly just adding the new data source, so I think the manual review should be good enough.

Sorry. I missed the tests in `dev` since they were disabled.
I just ran them, and verified that they pass (after fixing a issue with timezones pinned in them).

I will also add a few more tests for the new code.

ildar170975:
> A few things I think I will add

We should add warning messages in a console and ignore a “use statistics” option then.

ildar170975:
> This is also why the statistics option should definitely not be the default

Absolutely. It will be a breaking change we should not make.

src/buildConfig.js
```diff
@@ -221,6 +222,9 @@ export default (config) => {
         );
       }
     }
+    // `statistics` may be set per entity or card-wide, and accepts a bare
+    // `true` as shorthand for the defaults.
+    checkStatistics(conf, i);
```
ildar170975, line 227: Quick note: if some entity can be a card-wide too, it should be checked both per-enitity and card-wide. 

Update: I see, you fallback to the card-wide option in this function.

ildar170975:
Suggest to exclude words from Docs about “long-term”. Since you allow to select 5minute readings, these readings are not LTS. So, let’s just write “Statistics” in docs. 
Also, suggest to exclude these words from the PR title.

ildar170975:
About DEFAULT_STASTISTICS_TYPE. 
The “mean” type is only applicable for “measurement” state_class.
Theoretically, we can set a default type after analyzing a “state_class” attribute. 
If a manually selected type is not compatible - show a warning in console and fallback to a default type for that state_class.

Also, we can probably show a warning when the “statistics” option is set for an absent state_class (when stats are not supposed to be available). Away from a PC now, cannot check, earlier it still was possible to show statistics graph for OLD stored stats (if the state_class became “none” only recently).

Olen:
> Suggest to exclude words from Docs about “long-term”. Since you allow to select 5minute readings, these readings are not LTS. So, let’s just write “Statistics” in docs. Also, suggest to exclude these words from the PR title.

I see your point, and agree. I will change the terminology.

>  The “mean” type is only applicable for “measurement” state_class.

You are right.  
The statistics API has an endpoint that returns `has_mean`, `has_sum` etc. so we can use that to verify which types are available for an entity, and warn and fallback if the configuration does not make sense.  This data can be cached per card, so it only requires one extra call when the card is first loaded.
This also means that we can better support statistics that contain  `change` ("consumption over a period") as a parameter - typically energy, water or gas sensors.

HA has a dedicated "Repair" for when a state class is removed but the entity and metadata still exists. But of course removing the state_class means that no _new_ statistics will be generated for the entity.
I have also checked in my own DB that metadata survives even when the entity is completely gone.

Olen:
I added proper type-detection, along with `change` as a new type.
I tested the `list_statistic_ids` API endpoint, but there is no filter there, so it included _all_ statistics in each call.  That made it quite costly to do per card, even with caching.
It also does not describe the `measurement_angle` state_class correctly. The endpoint report `has_mean: false` while the statistics actually includes a `mean` value.

Instead I realized that the response we already fetch gives us this data for free.  The response includes the available types already, so we can just use that.  This also ensure that we allow only the types that each statistics actually provides data for.  Some keys exist in the response, but contain only `null` values, so we filter those out as well to avoid empty graphs.
If the user configures a type that is not available for the entity, we fall back to a default - `mean` if it is available, `state` otherwise. 

The only thing to be aware of is that we actually need a fetch of statistics to know what types are available for an entity.  To log any unavailable types, I also created a `logOnce()` helper, since it would otherwise log the wrong/missing type on every entity update.

PR #1430, opened 2026-08-20
fix: pass the built config to checkLineStyle

Olen:
Just a minor consistency fix.

Every other check in buildConfig receives `conf`.

`checkLineStyle()` receives the raw `config`.

No behaviour change today, but if `checkLineStyle` is extended (the name suggest that it could do more than just logging a warning) it would probably be a good idea to give it the normalized `conf` instead of the raw `config`.


ildar170975:
Absolutely!!!! Thanks!!!!

PR #1432, opened 2026-08-21
feat: Adaptive height

ildar170975:
Adaptive height of a graph when a card is placed inside a `horizontal-stack` or `grid` cards (and a sections layout which uses the `grid` internally) and a `height` option is not defined.
If the `height` option is defined - it is respected. Although this may cause a distortion: for example, when a mini-graph-card (MGC) becomes higher than a neighbouring card and this card DOES NOT support adaptive scaling.
Actually, same distortion may happen even with an undefined `height`: assume MGC is placed into a stack along with a very low card which does not scale itself, and MGC is automatically scaled down to a minimal height (100).

Two methods were considered:

1. Stretch/shrink a canvas. Managed to fix some distortions, failed to fix others.
2. Recreate a Graph object with a new height. After plenty of "try-fail-repeat", concluded that was the "truthy" way - so THIS method is implemented.

Both methods needed to track height changes & calculate a REQUIRED graph height (for method 2) or a CURRENT vertical scale (for method 1, was mainly needed to compensate distortions).

Important: now the card can adjust own height dynamically during a life cycle: for example, MGC is placed into a stack with a card with a dynamic height (like a vertical stack with cards with visibility conditions):

hhh

(animation is intentionally switched on, that is why the card is shown “not instantly”)

Notes:
1. Readme is corrected: a default value of `height` option is `100` (and has been at least from a creation of "buildConfig" module a few years ago) instead of `150`, let's consider this as a typo or a leftout from old versions.
2. This mentioned default value `100` now is only used when a card is NOT placed inside a `horizontal-stack` or `grid` cards (and a sections layout); otherwise this default value is used as a MINIMAL height.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ We recommend looking at the [Example usage section](#example-usage) to understan
 | cache | boolean | `true` | v0.9.0 | Enable/disable local caching of history data.
 | show | list |  | v0.2.0 | List of UI elements to display/hide, for available items see [available show options](#available-show-options).
 | animate | boolean | `false` | v0.2.0 | Add a reveal animation to the graph.
-| height | number | `150` | v0.0.1 | Set a custom height of the line graph.
+| height | number | `100` | v0.0.1 | Set a custom height of the line graph.
 | bar_spacing | number | `4` | v0.9.0 | Set the spacing between bars in bar graph. Value `-1` is used to place bars on each other. See [examples](#bar-spacing-examples).
 | bar_spacing_group | number |   | 0.14.0 | Set an additional spacing between bar groups (multiple entities) in bar graph. Fallback to `bar_spacing` if undefined; if `bar_spacing: -1` - then a default `4` value is used. See [examples](#bar-spacing-examples).
 | line_width | number | `5` | v0.0.1 | Set the thickness of the line.
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ We recommend looking at the [Example usage section](#example-usage) to understan
 | cache | boolean | `true` | v0.9.0 | Enable/disable local caching of history data.
 | show | list |  | v0.2.0 | List of UI elements to display/hide, for available items see [available show options](#available-show-options).
 | animate | boolean | `false` | v0.2.0 | Add a reveal animation to the graph.
-| height | number | `100` | v0.0.1 | Set a custom height of the line graph.
+| height | number | `100` | v0.0.1 | Set the height of the line graph in pixels.<br>If omitted, the height is calculated automatically (minimum `100`) when placed inside a `horizontal-stack` card, `grid` card, or sections view layout. Otherwise, it defaults to `100`. |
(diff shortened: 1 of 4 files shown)


ildar170975:
Some remarks for the alternative “scale canvas” method. 
When using a “circle” SVG command, after scaling  these circles become ellipses, so you need to de-scale them - for that you need to know a scale factor (as was said above) and a corresponding resize observer. 
Of course, you can draw a circle by lines (as was proposed in other PRs) instead.
But still there are other distortions which you cannot compensate. For example, lines may keep original thickness due to a corresponding style - that is OK. But margins of a graph area are also scaled; it is easier to see it for a bottom margin when no custom lower bound is defined and a graph has a flatline in some interval. This margin is supposed to be equal to a “line_width”, after up-scaling this margin becomes huge. 
That is why the “recreate Graph with a new height” method was chosen.

akloeckner:
Hey @ildar170975, I couldn't really make up my mind on what I think about the approach. I still want to offer my thoughts. Maybe it's helpful.

I think, it is even nice that the graph re-animates when sudden changes in the size happen. But what happens, if e.g. the window is resized continuously? Will the graph redraw all the time? That might be a showstopper.

I think, I would prefer to delegate all the monitoring and resizing to the browser. I have no working idea, but I would expect some "100% height" CSS directive, mixed with relative positioning of the drawn graph should be able to do that. 

I fixed the "circles become ellipses" in one of my grid-related PRs by drawing two points with different thicknesses on top of each other (#1199). But I also noticed the line thickness and margin issues in that PR. I didnt dig deeper at the time. You seem to have done some more research now. So, maybe it's just not feasible. 🤷 

Finally, I remember that the card should be fairly minimal. I haven't reviewed the code, but imagine it is quite a lot. So, it might become more difficult to maintain.

Looking back at my thoughts, it seems I don't like the approach by gut-feeling. I would give it another try to use some clever CSS. But if redrawing is the only feasible solution, you should go through with it. 👍  😉

ildar170975:
@akloeckner 
Hello! We can discuss in details in our chat in Community.

1. Unfortunately, I failed to solve the "margin distorted" problem with a "pure css" way. A canvas is scaled, margins are scaled too (((. Also. imagine that in future we can add more UI elements onto a canvas - which may be distorted too.
2. ResizeObserver added some more code to maintain. Yes, I would also prefer to keep the codebase as small as it could be to minimize a maintenance. But it is inevitable to have a codebase increased if we want to add more features ((. For instance, supporting native HA number/datetime format caused adding a new "locale.js" module.
3. There is a chance that a user will place the MGC card into a stack/grid with periodically resized card. Yes, a "showstopper" may happen - this will tell the user "do not do stupid things". A similar "showstopper" may happen if a user set "hours_to_show: 1000" & "points_per_hour: 2400" - we do not prevent this to happen and (may be) we should not. There is a common suggestion - "a card should adjust it's height automatically when placed in a stack", imho we need to provide this. Also - if a neighbouring card is tend to resize more than often - we can set a fixed height for MGC to avoid it's resizing.

The PR will be kept not merged, waiting for your proposals. Keep in touch!

Issue #983, opened 2023-07-07
Feature Request: Support profile time zone setting


tomlut:
HA Core 2023.7.0 introduced a profile setting for the frontend to display items in either the server or browser local time zone.

Would it be possible for the mini graph card to support this?

![Screenshot 2023-07-07 at 11-08-23 Profile – Home Assistant](https://github.com/kalkih/mini-graph-card/assets/10679300/1c367970-0d4a-4b24-8360-7df38fb2506f)

KM6IAU:
+1.  Although HA is missing full implementation of this new feature as of 2023.8.2, for instance in the range selector of the History view, I believe this FR would be the best default behavior for mini-graph-card.

ildar170975:
please check v0.14.0-dev.1

tomlut:
Sorry, I haven't used this card for years.

PR #1433, opened 2026-08-22
fix: respect Home Assistant time zone for graph buckets

rosenrot00:
Graph bucket boundaries now follow the time zone selected in the Home Assistant user profile.

History requests and graph aggregation use the same end time, including across daylight-saving transitions. Rolling interval graphs remain unchanged.


ildar170975:
Thank you, will check it. 
Seems to be related to - https://github.com/kalkih/mini-graph-card/issues/983

rosenrot00:
You are welcome. Yes, indeed this is related. I was surprised that it isn't implemented yet as it is annoying if you are on vacation and all mini graph cards are showing wrong values.

ildar170975:
Consider adding some tests. There is a corresponding test (disabled), check it in a “tests” folder.

rosenrot00:
@ildar170975 done.

ildar170975:
@rosenrot00 
Sorry for a delay, just got to a PC.

Question - why this PR is needed? 
Imho the `resolveTimeZone` function already supports the functionality you need.

Please check this:

1. Intentionally a "GMT+4" tz set on a PC, a server still uses "GMT+3":
image

2. When "use local tz" config settings is used:
image

Then a local "9:xx" time is shown:
image

3. When "use server time" is set:
image

Then a proper "8:xx" time is shown:
image

rosenrot00:
Thanks for checking.

`resolveTimeZone()` already handles timestamp formatting correctly, but it is currently not used when calculating graph bucket boundaries.

Both `MiniGraphCard.getEndDate()` and `Graph._updateEndTime()` still use browser-local `Date` setters. As a result, `group_by: date` and `group_by: hour` can request and aggregate a different period while displaying its labels in the time zone selected in Home Assistant.

The screenshots confirm that the displayed time is correct, but they do not verify the history request or aggregation boundaries. This PR addresses those boundaries and ensures that the same end time is used for both the history request and graph aggregation.

Rolling graphs using `group_by: interval` remain unchanged.

ildar170975:
> Both MiniGraphCard.getEndDate() and Graph._updateEndTime() still use browser-local Date setters. As a result, group_by: date and group_by: hour can request and aggregate a different period while displaying its labels in the time zone selected in Home Assistant.

I see!
Let me check it again.

PR #1457, opened 2026-09-14
fix: resolve entity names via hass.formatEntityName

bramkragten:
## What

Resolve the `name` option through `hass.formatEntityName` instead of reading `attributes.friendly_name` directly.

## Why

Home Assistant is moving from a single `friendly_name` string to a name composed from the entity's registry context (entity, device, area, floor). Under that model the device name becomes part of the composed name, so a card sitting in a device- or area-scoped dashboard repeats it — "Kitchen Sensor Temperature" where the built-in cards next to it show "Temperature". Cards that read the attribute directly also can't follow the default when HA changes it, and give users no way to choose which parts of the name they want.

Routing through the helper fixes both: the card matches the built-in cards, and users can compose the name themselves:

```yaml
entities:
  - entity: sensor.living_room_thermostat_temperature
    name:
      - type: area
      - type: entity
```

Context: home-assistant/frontend#51686, which tracks bringing the most-used custom cards along with the new naming model.

## Compatibility

Gated on HA 2026.4, where `formatEntityName` started accepting a card's `name` option (a user string, a structured name, or `undefined`). The helper has existed since 2025.10 but with an incompatible signature — it took bare type strings then, and 2025.11–2026.3 accepted only structured items — so feature detection isn't safe and the version is checked instead.

On older Home Assistant versions the behaviour is unchanged: a string `name` wins, otherwise `friendly_name`, otherwise the entity id. A structured `name` falls back to the friendly name there.

## Notes

- No new config option: the existing `name` accepts a list in addition to a string.
- `computeName` keeps its existing precedence, including the `Static` label for static values, which have no entity to resolve a name against.
- Documented in the README under "Structured names", with both `name` rows in the option tables pointing at it.
- Verified locally against `dev`: `npm run lint` passes clean and `npm run rollup` builds (the only warning is the pre-existing d3-interpolate circular-dependency notice). `dist/` and `package-lock.json` are untouched by this branch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


bramkragten:
Pushed a follow-up commit. `formatEntityName` resolves against the entity/device/area/floor registries, and Home Assistant swaps the real formatter in asynchronously once translations load — neither changes an entity's state object, so the existing update condition missed both. A device or area rename could leave a rendered name stale until an unrelated state change forced a render.

`shouldUpdate` now also compares `formatEntityName`, `entities`, `devices`, `areas` and `floors` by reference across `hass` assignments. Those are replaced wholesale rather than mutated, so a reference check is enough and stays cheap.

Credit to the Copilot review on denysdovhan/vacuum-card#1191, which spotted this in a sibling PR.

Re-verified: `npm run lint` clean and `npm run rollup` builds.

ildar170975:
Thanks Bram, will check.

PR #1467, opened 2026-09-20
fix: sort median aggregate values

dajiaohuang:
## What

Sort median aggregate samples by their numeric state before selecting the midpoint. The previous comparator parsed the sample objects themselves, leaving the input order unchanged and returning incorrect medians for unsorted history buckets.

## Validation

- `npx eslint src/* --ext .js --rule linebreak-style: off`
- `npm run rollup`
- focused numeric median regression for unsorted odd samples


ildar170975:
@dajiaohuang 
Well done!

